PR #79, opened 2014-06-07
Have subdomain sites work automatically

### LINKED ISSUE #42: Allow using multisite with subdomains
Here's how we can handle it:
- Drop vagrant-hostsupdater
- Enable avahi-daemon and [avahi-aliases](https://github.com/airtonix/avahi-aliases)
- Output the aliases in `/etc/avahi/aliases.d/default` via erb

Requires Bonjour on Windows hosts.

rmccue:
Changing host files sucks. Let's support this automatically with the magic of Avahi.

Fixes #42.


rmccue:
OK, @BronsonQuick tested this on macOS Sierra (10.12) and I tested on OS X Yosemite (10.10), and both work. Ready to merge!

Issue #114, opened 2014-12-17
Investigate alternative provisioning systems


rmccue:
Right now, Puppet is breaking our entire workflow by deprecating the `import` command. I have some workarounds we can probably apply here, but it's time we dig into alternative systems to see if we can get a better system. The fact that they're breaking this is a huge issue.

Top choice for an alternative right now is Salt, mainly because I need to know Salt for our Human Made configuration anyway.

rmccue:
I think we can get around this issue without switching away from Puppet, and alternative systems have issues. Closing.

PR #115, opened 2014-12-17
Remove Puppet imports

rmccue:
This converts from using Puppet's `import` to using symlinks into the main directory.

Should let us get away with keeping compatibility with existing extension format.

See #114, #109.


BronsonQuick:
I'm going to close this as it'll be taken care of in: https://github.com/Chassis/Chassis/pull/532/files

PR #127, opened 2015-04-24
Allow nginx to handle static assets

tareiking:
I had an issue where missing image assets would take a long time to return a 404. Turns out this was due to each request being handed off to fpm-php by nginx.

Allowing nginx to handle static assets made some useful performance gains, but not sure if this will have unintended effects - thoughts @rmccue?


rmccue:
Looks good to me, thanks @tareiking!

PR #129, opened 2015-04-24
Update wp 4.2

### LINKED ISSUE #128: Update to Wordpress 4.2
Can you please update WordPress to 4.2? Also, as I'm unclear how you are doing it, it would be awesome if you could write a few notes on how you update wp in Chassis, so in future I might be able to submit a pull request in stead.

It doesn't appear to be a submodule, so I'm a little unclear on how you do it.

I realise you might already be doing this, considering it's been out a few hours now. Thought it might be helpful to get some info on how you update though.

Thanks!

Sid

tareiking:
Fix #128


BronsonQuick:
Sorry. I was doing that on the call.

Issue #130, opened 2015-04-24
Use trusty64 instead precise64?


tikot:
What is the reason to use `precise64`?

I have recently tried to boot up Chassis could not get it to work. Changed box to `ubuntu/trusty64` boots up with out any problems. 

Should I submit a pull request to update box config to `ubuntu/trusty64` in Chassis/Vagrantfile line number 69

BronsonQuick:
Hey @tikot,

Can you please let me know what you couldn't get to work with `precise64`? Can you please uncomment: https://github.com/Chassis/Chassis/blob/master/Vagrantfile#L102 and post the output for me here please so we can work out what didn't work for you?

The main reason we have `precise64` is because when we started Chassis it was the last LTS release as well as the fact that it's the default Vagrant box some most people already have that base box on their machine.

That being said we are now keen to update to trusty too so a PR would be great thanks :+1: We'd still like to get to the bottom of your `precise64` issue too though! Thanks :smiley:

tikot:
Hey @BronsonQuick,

Here the log https://gist.github.com/tikot/4dc948e354df3a9f2598

BronsonQuick:
@rmccue Any thoughts on this one?

@tikot What version of Windows are you using? Also can you please let me know the versions of Vagrant and VirtualBox you're using? If they aren't the latest can you please try upgrading them?

It's a shame I left my Windows laptop in the office otherwise I could try and reproduce it. Thanks again! Hopefully we can get to the bottom of it!

tikot:
@BronsonQuick Latest version on everything.
Vagrant 1.7.2
VirtualBox 4.3.26r98988
Windows 8.1 x64

tikot:
Mac (log)[https://gist.github.com/tikot/e2ea8b91e879bef3d1fc] with php 5.5 config.yaml

Vagrant 1.7.2
VirtualBox 4.3.26r98988
OS X 10.10.3

I think I know what is the issue it is something with php https://github.com/Chassis/Chassis/pull/131#issuecomment-96450342

BronsonQuick:
I still haven't been able to replicate this one so I'll close it out.

PR #131, opened 2015-04-24
Update VM box to trusty64

tikot:
This update Chassis to use Trusty LTS as a VM box instead of Precise.

hopefully a fix for #130

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -66,7 +66,7 @@ Vagrant.configure("2") do |config|
 	vagrant_version = Vagrant::VERSION.sub(/^v/, '')
 
 	# We <3 Ubuntu LTS
-	config.vm.box = "hashicorp/precise64"
+	config.vm.box = "ubuntu/trusty64"
 
 	# Having access would be nice.
 	if CONF['ip'] == "dhcp"


BronsonQuick:
Thanks @tikot :+1: I'm just testing this out now with all the various configurations and Chassis extensions. I'll add a list of things I'm checking and I'll tick them off as I go :smile: 

- [ ] Php 5.3
- [ ] Php 5.4
- [ ] Php 5.5
- [ ] Php 5.6
- [ ] https://github.com/Chassis/phpMyAdmin
- [ ] https://github.com/Chassis/memcache
- [ ] https://github.com/Chassis/Tester
- [ ] https://github.com/Chassis/mailcatcher

BronsonQuick:
Dang. Php 5.6.7-1 is installed during the first `vagrant up` which means our php version in `config.yaml` is being ignored :cry: 

Changing the php version in `config.yaml` to 5.5 breaks Chassis as well. Looks like I might need a hack day to implement this update without breaking anything. Sorry @tikot :( At least you can run Trusty on your machine in the meantime I guess!

tikot:
its never that simple :)

the php 5.3 & 5.4 is not supported in Ubuntu 13 and higher [ppa:ondrej/php5-oldstable](https://launchpad.net/~ondrej/+archive/ubuntu/php5-oldstable)

And by adding this [php.pp#L7] (https://github.com/Chassis/Chassis/blob/master/puppet/modules/sennza/manifests/php.pp#L7) it installs php 5.6.7-1

PR #140, opened 2015-06-16
Map extra directories into the VM

kpullen:
I've got a Wordpress theme that I'd like to keep in a repository with
some non-Wordpress stuff. I want to use Chassis during development, but
I don't want to muck about with Chassis in my repository as well. I can
symlink `Chassis/content` or bits inside it all day long, but the VM
that Chassis creates has no idea what is on the other end of that
symlink.

So, I've added a tiny bit to the Vagrantfile that searches through the
coalesced config for an entry named `synced_folders`. It'll go through
the entire list, mapping the key (on the host filesystem) to the value
(on the virtual filesystem).

Am I insane? I feel like not, but you know, let me know. Should this be an extension? If so, how do you suggest I build it?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -120,8 +120,14 @@ Vagrant.configure("2") do |config|
 	# Ensure that WordPress can install/update plugins, themes and core
 	if vagrant_version >= "1.3.0"
 		config.vm.synced_folder ".", "/vagrant", :mount_options => [ "dmode=777,fmode=777" ]
+		CONF["synced_folders"].each do |from, to|
+			config.vm.synced_folder from, to, :mount_options => [ "dmode=777,fmode=777" ]
+		end if CONF["synced_folders"]
 	else
 		config.vm.synced_folder ".", "/vagrant", :extra => "dmode=777,fmode=777"
+		CONF["synced_folders"].each do |from, to|
+			config.vm.synced_folder from, to, :extra => "dmode=777,fmode=777"
+		end if CONF["synced_folders"]
 	end
 
 	# Success?
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -205,3 +205,20 @@ instead, such as:
 .. code-block:: yaml
 
    apt_mirror: http://mirror.optus.net/ubuntu/
+
+
+Synced Folders
+--------------
+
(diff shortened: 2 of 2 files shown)


Japh:
@kpullen Any reason why you wouldn't use a structure something like the following:

```
+ chassis (Chassis repo)
|----+ content (Project repo)
      |----+ themes
            |----+ your-theme (submodule)
```

That way the theme is in it's own separate repo, but you needn't mess about with symlinks etc.

kpullen:
Good question! You can ignore that thing I said about symlinks. I was simply explaining that I tried to map the directories in with symlinks because I (in usual idiotic fashion) didn't realize that the moment Chassis was mapped into the VM the symlink would be pointing to a place on the virtual filesystem that probably didn't exist. No symlinks would really be necessary to add or use this feature.

For my current project I want to have the following structure:

```
+ root
|--+ Chassis (submodule)
|--+ My theme
|--+ Non-wordpress stuff
```

As you can see, I'd really like to have a single repository containing my theme and the other non-wordpress code, with Chassis sitting next to everything so another developer can just clone this repo, run a setup script (2 lines, I think) and be on their way. With the traditional approach I'll either need to:

  1. Get every developer to download Chassis and then add our repo as a submodule.
  2. Include Chassis in the project and keep our theme deeply nested inside Chassis.

I think the first option is possible, but annoying. It also puts a bunch of non-wordpress theme code in the themes directory (probably not really a problem, but it's kind of ugly to me). I think the second option mixes project code with what I consider "tooling" code.

I guess in the end I really want to treat Chassis as a tool, and not have my developers make work-arounds or munge the repo too much to fit on Chassis. What do you think?

_NB_ The setup script I described above is totally unnecessary for most projects, and no one needs a setup script to use this feature.

rmccue:
So, the best way to do this is to reverse symlink. Clone your theme into Chassis/content/themes, then symlink from the top-level into the Chassis dir.

kpullen:
@rmccue I'm sorry, I don't fully understand what you mean. Could you clarify a little?

rmccue:
Oh gosh, I was totally wrong on this :)

Loving this PR, thank you!

PR #145, opened 2015-08-24
Support arbitrary paths for WP & content

rmccue:
Right now, we only support having `content` and `wp` under the Chassis root directory. This is problematic if you want to use Chassis with an existing project.

Let's fix that. ;)

This PR adds a `paths` configuration option, which lets you point `base`, `content` and `wp` at anything you'd like. Both `content` and `wp` are taken as relative to `base`, but can independently be set to other absolute paths.

Here's how we process it:
1. If `base` is not somewhere under Chassis' directory, mount it as a shared folder at `/chassis`
2. If `content` is not somewhere under `base`, mount it as a shared folder at `/chassis/content`
3. If `wp` is not somewhere under `base`, mount it as a shared folder at `/chassis/wp`

There's a potential conflict if your base already has a `content` or `wp` and you're using another one, but that's a pretty edge-case kinda thing.

(Don't merge yet, needs more testing and docs. :) )

Testing required:
* [x] Base: default, WP: default, content: default
* [x] Base: default, WP: default, content: non-default (under base)
* [x] Base: default, WP: default, content: non-default (not under base)
* [ ] Base: default, WP: non-default (under base), content: default
* [x] Base: default, WP: non-default (not under base), content: default
* [ ] Base: default, WP: non-default, content: non-default
* [x] Base: non-default (others must be non-default)


rmccue:
Non-default WP requires changing the `index.php` right now, so that's annoying.

rmccue:
The logic here is quite complex, juggling between the defaults and the customised version. This means that you can never be sure where WP is (`/vagrant/wp`, `/chassis/wp`, or some other random location if it's under the root), nor the content.

Considering rewriting this to always use `/chassis` instead, but I think that might end up making it more brittle...

PR #146, opened 2015-08-26
Avoid referring directly to the latest WordPress version number

johnbillion:
This way it doesn't go stale.


rmccue:
:sparkling_heart:

Issue #20, opened 2013-10-28
Add and configure OpenSSL


BronsonQuick:
Looks like https://github.com/camptocamp/puppet-openssl might do the trick although you might know a better way @rmccue

javorszky:
Hey, so I created this: https://github.com/javorszky/chassis-openssl

It works, but is brittle. If you have some time, can you look at that / help me what I can improve and how? :)

javorszky:
So I was thinking about this and have been trying to get it working. Findings: 

* simply overwriting the nginx file in `sites-available` is not an option, as puppet will complain that the same file is declared someplace else
* having a nearly identical nginx file in the same folder for what is effectively 3 lines of change for the entire functionality seems a bit of an overkill, plus the extension would need to closely monitor chassis core and adjust its own template files as core's change
* instead, it might be an option to add this into core, and put this behind a `yaml` flag: `-ssl: on`, and much like the PHP versions, register it, but don't pull it down, unless the user has specified it
* that way there could be an alternate template, and provisioning could choose which one to use whether the flag is set in the yaml file

Thoughts? :)

rmccue:
@javorszky I'd like to keep it out of Chassis by default, since it doesn't fit the 80/20 rule, plus I think we're better off making a more generic system instead.

What if instead we added `include /etc/nginx/sites-enabled/{$fqdn}.d/*` into there? The SSL extension could then just have `{$fqdn}.d/ssl` with:
```nginx
listen 443 ssl;

ssl_certificate    /etc/ssl/certs/<%= @fqdn %>.crt;
ssl_certificate_key    /etc/ssl/certs/<%= @fqdn %>.key;
```

I'm investigating if this would work at the moment. :)

rmccue:
I just tried this locally, and it seems like you can do this:

```nginx
listen 80;
listen 8080;
listen 443 ssl;

ssl_certificate    test.crt;
ssl_certificate_key    test.key;
```

...all in the same server block, and it works fine. The SSL configuration only applies to the listening ports where `ssl` is specified.

@javorszky Would this work for you?

javorszky:
Hey @rmccue!

Awesome, thank you! TIL about nginx's include.

The above description didn't work for me though. It gets stuck while trying to include `/etc/nginx/sites-enabled/*`, and sadly `{$fqdn}.d/` is also being picked up, but is a directory, and nginx doesn't seem to like that.

Working on it though...

BronsonQuick:
[...]

PR #150, opened 2015-09-16
Add line to include extra nginx directives into site specific conf file

javorszky:
Connected to #20

Why include from within sites-available?

* if the files are in sites-enabled/server.d/*, then the general nginx conf will try to include server.d as well, and will fail, as it's a directory
* since the extra files are fragments of nginx directives, including them on sites-enabled/ will break nginx (missing curly braces, etc)
* extra files could be therefore anywhere, and sites-available is as good as any as long as directories there aren't symlinked to sites-enabled

### DIFF
--- a/puppet/modules/sennza/templates/site.nginx.conf.erb
+++ b/puppet/modules/sennza/templates/site.nginx.conf.erb
@@ -4,6 +4,8 @@ server {
 	root <%= @location %>;
 	client_max_body_size 50M;
 
+	include /etc/nginx/sites-available/<%= @fqdn %>.d/*;
+
 	location / {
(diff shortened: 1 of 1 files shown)


javorszky:
Also allowed me to implement these changes: https://github.com/javorszky/chassis-openssl/commit/47a5a7fe87cb1a446190d2db6a24c62189115b3b :)

rmccue:
This is exactly what I was thinking; thanks! :)

BronsonQuick:
Nice work @javorszky :smile: I was on holidays for a few weeks so I'm only just catching up on your pull requests. Taa for the merge @rmccue!

PR #153, opened 2015-10-16
Add ssh forwarding by default

terminalpixel:
Just a simple little thing to have that makes deploying potentially far more simple.

There may be a reason for not including this by default but I'm not aware of it.


rmccue:
Seems like a good idea to me! Can you please remove the trailing whitespace on the blank lines though? :)

Thanks for the PR!

terminalpixel:
Oops, updated that now. Let me know if you want it merged into a single commit or not

rmccue:
Nope, separate is fine. Thanks again!

PR #188, opened 2016-06-22
Update to WordPress 4.5.3

lukecarbis:
Question: Couldn't Chassis fetch the latest WordPress version from Github when it provisions? It could clone / pull the latest changes into a seperate repo, which is `.gitignore`d?


BronsonQuick:
@lukecarbis Thanks for updating the subtree!

Answer: I guess we could do that but the main reason behind using subtrees is to reduce the size of the initial clone. Were you thinking submodules to do that or a bash script?

lukecarbis:
That makes sense. I was thinking a bash script. You're right, unless we downloaded and unzip `latest`, cloning the whole repo would take much longer.

BronsonQuick:
@lukecarbis Thanks for the PR. Normally I'm onto it straight away but I was/still am in transit!

Issue #189, opened 2016-06-27
Docs: Remove reference to Supercharger


peterwilsoncc:
Remove supercharger reference in docs at http://docs.chassis.io/en/latest/quickstart/ and elsewhere.

PR #192, opened 2016-06-27
Remove supercharger reference from docs

dan-westall:
#189 - Remove supercharger reference from quick-start docs as project is no longer actively maintained


BronsonQuick:
@dan-westall Thanks! 😍

PR #207, opened 2016-07-18
HTTPS support

johnbillion:
This is currently a WIP.

I'm stuck due to the following error which occurs during provisioning, and it's rather short on details (no file or line number):

```
==> default: Error: Could not run: no such file to load -- inifile
```

The OpenSSL module has two references to `inifile`:
1. https://github.com/camptocamp/puppet-openssl/blob/2fdc822663248265ccb6c5bb6a8b498ada67dc49/lib/puppet/provider/x509_cert/openssl.rb#L39
2. https://github.com/camptocamp/puppet-openssl/blob/4add0969ac876a9d3976540cdf77d9838f37d829/spec/unit/puppet/provider/x509_cert/openssl_spec.rb#L1

I'm guessing that one of these is causing a problem but I'm unsure how to go about debugging it.

Any help, suggestions welcome.

### DIFF
--- a/puppet/modules/chassis/manifests/openssl.pp
+++ b/puppet/modules/chassis/manifests/openssl.pp
@@ -1,0 +1,28 @@
+define chassis::openssl (
+) {
+	file { '/vagrant/certs':
+		ensure => 'directory',
+	}
+
+	# include ::openssl
+
+	# class { '::openssl':
+	  # package_ensure         => latest,
+	  # ca_certificates_ensure => latest,
+	# }
+	
+	$commonname = "*.${fqdn}"
+
+	openssl::certificate::x509 { $fqdn:
+	  country      => 'HM',
+	  organization => $fqdn,
+	  commonname   => $commonname,
+	} ->
+	file { "/vagrant/certs/${fqdn}.crt":
+		ensure => 'present',
+		source => "/etc/ssl/certs/${fqdn}.crt",
+		mode => '0644',
+	}
+
+	include ::openssl-nginx
+}
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -41,4 +41,8 @@ define chassis::site (
 		admin_email    => $admin_email,
 		admin_password => $admin_password,
 	}
+
+	chassis::openssl { $name:
+	}
+
 }
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -10,6 +10,7 @@ define chassis::site (
 	$admin_user     = 'admin',
 	$admin_email    = 'admin@example.com',
 	$admin_password = 'password',
+	$https          = false,
 ) {
 	$extra_hosts = join($hosts, ' ')
 	$server_name = rstrip("${name} ${extra_hosts}")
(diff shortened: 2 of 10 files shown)


puppet/modules/chassis/manifests/openssl.pp
```diff
@@ -0,0 +1,28 @@
...
+	  # package_ensure         => latest,
+	  # ca_certificates_ensure => latest,
+	# }
+	
+	$commonname = "*.${fqdn}"
+
+	openssl::certificate::x509 { $fqdn:
+	  country      => 'HM',
+	  organization => $fqdn,
+	  commonname   => $commonname,
```
rmccue: These need to use tabs, rather than spaces.

puppet/modules/chassis/manifests/openssl.pp
```diff
@@ -0,0 +1,28 @@
...
+
+	# class { '::openssl':
+	  # package_ensure         => latest,
+	  # ca_certificates_ensure => latest,
+	# }
+	
+	$commonname = "*.${fqdn}"
+
+	openssl::certificate::x509 { $fqdn:
+	  country      => 'HM',
```
rmccue: I'd rather we use a [private code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#User-assigned_code_elements) here; `AA` or `ZZ` sound preferable.

puppet/modules/chassis/manifests/openssl.pp
```diff
@@ -0,0 +1,28 @@
...
+	  organization => $fqdn,
+	  commonname   => $commonname,
+	} ->
```
rmccue: This should be specified as `require => Openssl::Certificate::X509[$fqdn]` on the file resource instead.

puppet/modules/chassis/manifests/openssl.pp
```diff
@@ -0,0 +1,28 @@
+define chassis::openssl (
+) {
+	file { '/vagrant/certs':
```
rmccue: Not sure why this needs to provision files/dirs in the vagrant directory; is the idea that this allows easier access to install on the host?

puppet/modules/chassis/manifests/site.pp
```diff
@@ -34,11 +41,15 @@
...
+		url => $url,
 		name => 'Vagrant Site',
 		require => Mysql::Db[$database],
 		admin_user     => $admin_user,
 		admin_email    => $admin_email,
 		admin_password => $admin_password,
 	}
+
+	chassis::openssl { $name:
+	}
```
rmccue: I'm assuming this should also `require` the nginx provisioning?

puppet/modules/openssl-nginx/manifests/init.pp
```diff
@@ -0,0 +1,12 @@
+class openssl-nginx {
+	file { "/etc/nginx/sites-available/$fqdn.d":
+		ensure => directory,
+		require => Package['nginx']
+	}
```
rmccue: We should probably have this one in our regular nginx provisioning.

johnbillion:
I'm going to close this for now as chassis-openssl does the job. I might reopen it when I get some spare time (HAHAHAHA).

PR #209, opened 2016-07-23
Add wp-cli config in homedir for easier calls from outside vagrant

### LINKED ISSUE #204: Link wp-cli.yml to home folder
This will allow direct usage of wp-cli using `vagrant ssh -c 'wp foo bar'` commands, without the need to SSH then issue a single command.

shadyvb:
Resolves #204


puppet/modules/chassis/manifests/wp.pp
```diff
@@ -56,4 +56,8 @@
 	file { '/vagrant/local-config-extensions.php':
 		content => template('chassis/local-config-extensions.php.erb')
 	}
+
+	file { '/home/vagrant/.wp-cli/config.yml':
+		content => template('chassis/wp-cli.yml')
```
BronsonQuick: @shadyvb Could you please change this to be `chassis/wp-cli.yml.erb` please and also rename the the file accordingly as well? Thank you! ❤️

shadyvb: @BronsonQuick Just did 👍

BronsonQuick:
@shadyvb Thank you Sir! 👍

PR #210, opened 2016-07-24
Add config section to install/activate plugins and a default theme

### LINKED ISSUE #203: Allow setting default theme/plugins list from config file
It'd be nice if we can setup the default theme, and the active/installed plugin list through the config.yaml file, would make it easier for people to have a consistent environment across teams.

shadyvb:
Resolves #203

### DIFF
--- a/puppet/modules/chassis/manifests/wp.pp
+++ b/puppet/modules/chassis/manifests/wp.pp
@@ -65,11 +65,11 @@ define chassis::wp (
 
 	wp::plugin { $plugins:
 		location => $location,
-		ensure => 'enabled',
+		ensure   => 'enabled',
 	}
 
 	wp::theme { $themes:
 		location => $location,
-		ensure => 'enabled',
+		ensure   => 'enabled',
 	}
 }
--- a/puppet/modules/chassis/manifests/wp.pp
+++ b/puppet/modules/chassis/manifests/wp.pp
@@ -12,7 +12,7 @@ define chassis::wp (
 	$admin_password = 'password',
 	$network = false,
 	$plugins = [],
-	$theme = '',
+	$themes = [],
 
 	$extensions = [],
 ) {
@@ -68,10 +68,8 @@ define chassis::wp (
 		ensure => 'enabled',
 	}
 
-	if ( $theme ) {
-		wp::theme { $theme:
-			location => $location,
-			ensure => 'enabled',
-		}
+	wp::theme { $themes:
+		location => $location,
+		ensure => 'enabled',
 	}
 }
--- a/puppet/modules/chassis/manifests/wp.pp
+++ b/puppet/modules/chassis/manifests/wp.pp
@@ -11,6 +11,8 @@ define chassis::wp (
 	$admin_email    = 'admin@example.com',
 	$admin_password = 'password',
 	$network = false,
+	$plugins = [],
+	$theme = '',
 
 	$extensions = [],
 ) {
@@ -59,5 +61,17 @@ define chassis::wp (
 
 	file { '/home/vagrant/.wp-cli/config.yml':
 		content => template('chassis/wp-cli.yml.erb')
+	}
+
+	wp::plugin { $plugins:
+		location => $location,
+		ensure => 'enabled',
+	}
+
+	if ( $theme ) {
+		wp::theme { $theme:
+			location => $location,
+			ensure => 'enabled',
+		}
 	}
 }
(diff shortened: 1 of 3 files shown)


puppet/modules/chassis/manifests/wp.pp
```diff
@@ -60,4 +62,16 @@
 	file { '/home/vagrant/.wp-cli/config.yml':
 		content => template('chassis/wp-cli.yml.erb')
 	}
+
+	wp::plugin { $plugins:
+		location => $location,
+		ensure => 'enabled',
```
BronsonQuick: Reckon you could line up the `=>` please? 👍

puppet/modules/chassis/manifests/wp.pp
```diff
@@ -11,6 +11,8 @@
 	$admin_email    = 'admin@example.com',
 	$admin_password = 'password',
 	$network = false,
+	$plugins = [],
+	$theme = '',
```
BronsonQuick: I'm thinking it might be good to do an array of themes as well? The rationale behind that is that sometimes with multisite you'll have a few different themes and not just one. We could perhaps make it so the first of the themes listed in the `.yaml` file is `enabled`. What are your thoughts Sir?

BronsonQuick:
@shadyvb Yay! Thanks for your PR's. It's great to have someone else excited about Chassis too ❤️

shadyvb:
Couple of problems with this which happens on subsequent provisioning:

- Plugins are activated on each provision, should only install/activate on the first time ?
- Theme will always be activated on each provision

And a relevant note:

- Plugins cannot just be installed without activation, limitation of puppet-wp, though not really a big issue, can't think of many scenarios where one would want to install a plugin but not activate it

puppet/modules/chassis/manifests/wp.pp
shadyvb: Unless you're talking about installing wp.org themes ( pretty rare I guess with such setup ), you wouldn't really need to do this since you'd be cloning / sub modeling them within content folder. Do you think we should cater for this edgy scenario ?

BronsonQuick:
>Plugins cannot just be installed without activation, limitation of puppet-wp, though not really a big issue, can't think of many scenarios where one would want to install a plugin but not activate it

Ahh dang. I wasn't aware that was an issue 😞  We might have to patch that upstream to make this more robust?

shadyvb:
Yeah, I just can't think of a solution where I don't use a DB option to track this one :/ Not a puppet pro to know if there is a solution for that using puppet itself.

shadyvb:
Sorry, I meant re: the subsequent provisions, that bit about plugins installed and not activated can sure be patched upstream, though not a biggie to handle for this issue at hand, pretty edgy scenario.

puppet/modules/chassis/manifests/wp.pp
BronsonQuick: @shadyvb I think we should make it handle multiple wp.org themes. I know internally we'll rarely ever need to do that as we'll have submodules but I think other folks using Chassis might end up needing multiple themes from wp.org. Sorry to make it more difficult for you but we always try and think about all the use cases for Chassis 👍

BronsonQuick:
> Yeah, I just can't think of a solution where I don't use a DB option to track this one :/ Not a puppet pro to know if there is a solution for that using puppet itself.

We could perhaps just check for the existence of the theme(s)/plugin(s) directory/slug and if that exists assume it's been installed? I'd jump in and help you now but I'm in the middle of coding up stuff for my WordCamp Sydney talk 😁

shadyvb:
Raised https://github.com/rmccue/puppet-wp/issues/33 for the mentioned issue, might help with the subsequent provisions thing for some cases, where the plugin isn't already installed within the content directory. But will still cause a problem if you're only activating a cloned/submodules plugin.

shadyvb:
@BronsonQuick np buddy, I'm on it.

shadyvb:
Since we cannot install but not activate a theme, this will cause all themes to be activated, and might trigger unneeded hooks. Will probably need the upstream patch to remedy this side-effect.

puppet/modules/chassis/manifests/wp.pp
```diff
@@ -68,10 +68,8 @@
...
 	}
 
-	if ( $theme ) {
-		wp::theme { $theme:
-			location => $location,
-			ensure => 'enabled',
-		}
+	wp::theme { $themes:
+		location => $location,
+		ensure => 'enabled',
```
BronsonQuick: Reckon you could add another space to line up the `=>` again. Sorry to be so anal about it 😄

PR #225, opened 2016-08-05
Install extensions from config #55

### LINKED ISSUE #55: Autoinstall extensions from Git
We should allow specifying an `extensions` option in config.yaml that will autoinstall extensions correctly.

```yaml
# Autoinstall the following extensions
# (Note that all directories in extensions will be loaded; this is only for initial provisioning)
extensions:
  memcache: Chassis/memcache # mapped to https://github.com/Chassis/memcache.git
  xhprof: https://github.com/Chassis/xhprof.git
```

shadyvb:
Resolves #55

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -85,4 +85,20 @@ module Chassis
 	def self.config
 		self.normalize_config(self.load_config())
 	end
+
+	def self.install_extensions(config)
+	    # Install extensions listed in config
+        if config["extensions"]
+            config["extensions"].each do |ext|
+                begin
+                    repo = /^(git@|https:)/.match(ext) ? ext : 'https://github.com/' + ext
+                    folder = ext.split('/').last.gsub(/.git$/, '')
+
+                    if ! File.exist?( @@dir + "/extensions/" + folder )
+                        system("git clone %s %s/extensions/%s --recursive" % [repo, @@dir, folder] )
+                    end
+                end
+            end
+        end
+    end
 end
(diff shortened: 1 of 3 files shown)


puppet/chassis.rb
```diff
@@ -85,4 +85,20 @@ def self.normalize_config(config)
 	def self.config
 		self.normalize_config(self.load_config())
 	end
+
+	def self.install_extensions(config)
+	    # Install extensions listed in config
+        if config["extensions"]
```
rmccue: Looks like this needs spaces -> tabs

puppet/chassis.rb
```diff
@@ -85,4 +85,20 @@ def self.normalize_config(config)
...
+	def self.install_extensions(config)
+		# Install extensions listed in config
+		if config["extensions"]
+			config["extensions"].each do |ext|
+				begin
+					repo = /^(git@|https:)/.match(ext) ? ext : 'https://github.com/' + ext
+					folder = @@dir + '/extensions/' + ext.split('/').last.gsub(/.git$/, '')
+
+					if ! File.exist?( folder )
+						system("git clone %s %s --recursive" % [repo, folder] )
```
rmccue: Does this output to the terminal?

shadyvb: Yup

rmccue: neato!

rmccue:
✨ 💯 🌴

BronsonQuick:
💥

PR #246, opened 2016-11-01
Map the nginx and php logs

### LINKED ISSUE #245: symlink to log file
can we get a symlink to the web server or php log files in the user's home folder? I can never find where it lives and it seems different on different linuxes, so this would be convenient

BronsonQuick:
Fixes #245.

@rmccue I'm curious to get your thoughts on this. I needed to add in `:create => true` to make sure that folders were created on the host machine.


rmccue:
> I needed to add in :create => true to make sure that folders were created on the host machine.

I think this is only required if the local directories don't exist?

BronsonQuick:
>I think this is only required if the local directories don't exist?

Correct! I wanted them to be created on provision if they didn't exist.

BronsonQuick:
@rmccue Cheers for the feedback. I've updated the PR.

PR #248, opened 2016-11-01
Updating comments regarding the use of subdomain mutisite

cameronjonesweb:
In the docs it specifies that using subdomains for multisite is possible (http://docs.chassis.io/en/latest/config/#multisite), but the comments in config.yaml say it's not supported  yet. I'm going to assume the docs are correct and that the comments just need to be updated

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -38,7 +38,7 @@ ip: dhcp
 apt_mirror: Yes
 
 # Should we use multisite?
-# (Set to Yes for subdorectories)
+# (Set to Yes for subdirectories)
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -38,8 +38,8 @@ ip: dhcp
 apt_mirror: Yes
 
 # Should we use multisite?
-# (Only subdirectories are supported currently)
-# Values: Yes, No
+# (Set to Yes for subdorectories)
```
BronsonQuick: Could you please change your `subdorectories` typo to `subdirectories` Sir?

PR #257, opened 2016-11-09
Update ppas

### LINKED ISSUE #160: Add Php7
The apt is `ppa:ondrej/php` but  I think we'll need to do more configuration in https://github.com/Chassis/Chassis/blob/master/puppet/modules/sennza/manifests/php.pp

### LINKED ISSUE #256: Change the ppa's for PHP
The following ppa's are no longer valid:
```
apt::ppa { "ppa:ondrej/php5": }
apt::ppa { "ppa:ondrej/php5-5.6": }
```
We need to change to:
`apt::ppa { "ppa:ondrej/php": }`

BronsonQuick:
The ppa's for php recently changed so this PR updates them accordingly to fix #256.

As a bonus this also fixes #160.

NB: This PR hasn't accounted for version switching after an initial provision as I'll open a new issue for that. Currently with the ppa's we have at the moment Chassis is broken so merging this is more important that version switching after an initial provision.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -35,7 +35,7 @@ ip: dhcp
 # built-in repositories, or a mirror domain as a string.
 #
 # Values: Yes, No, or any mirror domain name e.g. http://mirror.optus.net/ubuntu/
-apt_mirror: Yes
+apt_mirror: No
 
 # Should we use multisite?
 # (Set to Yes for subdirectories)
--- a/config.yaml
+++ b/config.yaml
@@ -98,5 +98,5 @@ synced_folders:
 #  - git@github.com:Chassis/memcache.git
 
 # PHP version
-# Values: 5.3, 5.4, 5.5, 5.6 (or e.g. 5.3.3)
+# Values: 5.3, 5.4, 5.5, 5.6, 7.0, 7.1 (or e.g. 5.3.3)
 php: 5.6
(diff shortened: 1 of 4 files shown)


BronsonQuick:
I should also note I've tested initial provisions with the following versions of PHP:
- [x] 5.3
- [x] 5.4
- [x] 5.5
- [x] 5.6
- [x] 7.0
- [x] 7.1

config.yaml
```diff
@@ -35,7 +35,7 @@ ip: dhcp
 # built-in repositories, or a mirror domain as a string.
 #
 # Values: Yes, No, or any mirror domain name e.g. http://mirror.optus.net/ubuntu/
-apt_mirror: Yes
+apt_mirror: No
```
rmccue: This seems like it should be in a different PR?

puppet/manifests/development.pp
```diff
@@ -5,8 +5,6 @@
 $extensions = sz_extensions('/vagrant/extensions')
 $php_extensions = [ 'curl', 'gd', 'mysql' ]
 
-Class['mysql'] -> Package['php5-mysql']
-
```
rmccue: I think we still need this dependency; should it be moved into php.pp as `${php_package}-mysql`?

BronsonQuick: Nah we don't need this here anymore as I handled it here instead: https://github.com/Chassis/Chassis/pull/257/files#diff-ebb1484b5498abe97a80e94aac9365cdR58

BronsonQuick:
@rmccue Back to you. Should be good to merge now and get Chassis working again.

I'll work on https://github.com/Chassis/Chassis/issues/259

BronsonQuick:
Merging. Sorry @rmccue I need this in for one of our client projects!

PR #264, opened 2016-11-25
Add .vagrantplugins to empower extensions

rmccue:
Since Vagrant 1.7.0, a `.vagrantplugins` file [has been supported](https://github.com/mitchellh/vagrant/issues/3775). This allows Vagrant boxes to specify any arbitrary Ruby code to be loaded; in particular, Vagrant plugins and custom commands.

While we don't have any uses for this ourselves yet, we should allow extensions to define this.

This code finds any `.vagrantplugins` files in extensions, and loads them in from our primary file.

### DIFF
--- a/.vagrantplugins
+++ b/.vagrantplugins
@@ -1,13 +1,3 @@
-#puts "loaded"
-
-# class Command < Vagrant.plugin("2", :command)
-# 	def execute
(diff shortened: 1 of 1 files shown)


rmccue:
@BronsonQuick to you for review

Issue #268, opened 2016-12-01
Investigate "could not set 'file' on ensure" issues


rmccue:
I've heard this one a couple of times now. Let's try and track it down:

```
==> default: Error: Could not set 'file' on ensure: No such file or directory - /etc/nginx/sites-available/hm-vip.local20161201-2324-iydb4v.lock at 22:/vagrant/puppet/modules/chassis/manifests/network.pp
==> default:
==> default: Error: Could not set 'file' on ensure: No such file or directory - /etc/nginx/sites-available/hm-vip.local20161201-2324-iydb4v.lock at 22:/vagrant/puppet/modules/chassis/manifests/network.pp
==> default: Wrapped exception:
==> default: No such file or directory - /etc/nginx/sites-available/hm-vip.local20161201-2324-iydb4v.lock
==> default: Error: /Stage[main]/Main/Chassis::Wp[hm-vip.local]/Chassis::Network[hm-vip.local]/File[/etc/nginx/sites-available/hm-vip.local]/ensure: change from absent to file failed: Could not set 'file' on ensure: No such file or directory - /etc/nginx/sites-available/hm-vip.local20161201-2324-iydb4v.lock at 22:/vagrant/puppet/modules/chassis/manifests/network.pp
```

rmccue:
Seems to be fixed per @mattheu's testing, but reopen this if it pops back up again.

PR #270, opened 2016-12-01
Ensure the sites-available/enabled directories exist

rmccue:
I have a theory about #268: nginx is installed after we try and create the config files. This could happen since we don't have (or need) dependencies between nginx and everything else.

To resolve it, I've added an ensure for the directories. On already working installs, this will be a no-op, as the directories already exist. Puppet will create an implicit dependency between the files and the directory, so this should be fine.

One potential problem is that `/etc/nginx` may not exist, and this isn't in recursive mode. To investigate further.

### DIFF
--- a/puppet/modules/chassis/manifests/init.pp
+++ b/puppet/modules/chassis/manifests/init.pp
@@ -23,13 +23,13 @@ class chassis {
 		require    => Package['nginx']
 	}
 
+	file {['/etc/nginx', '/etc/nginx/sites-available', '/etc/nginx/sites-enabled']:
+		ensure => directory
+	}
+
 	file {'/etc/nginx/nginx.conf':
 		content => template('chassis/nginx.conf.erb'),
 		require => Package['nginx'],
 		notify => Service['nginx']
-	}
-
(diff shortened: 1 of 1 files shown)


rmccue:
So, as it turns out, the config files _do_ rely on the nginx service, since they `notify` it. However, my guess is that nginx isn't creating these dirs for some reason?

I can reliably recreate the issue by commenting the `package {'nginx':}` and related dependencies, although this causes an extra failure since `/etc/nginx` doesn't exist when we create the main config file. Updated the PR to add this one too, which ensures everything succeeds when the package is disabled out.

@BronsonQuick #reviewmerge

puppet/modules/chassis/manifests/init.pp
```diff
@@ -23,6 +23,10 @@
 		require    => Package['nginx']
 	}
 
+	file {['/etc/nginx', '/etc/nginx/sites-available', '/etc/nginx/sites-enabled']:
```
BronsonQuick: I think we need to work out some coding standards for our Puppet files. I kinda want to see more spaces inside '{' and '}'. That's not a blocker for this just something to note so we can work on that going forward.

Issue #193, opened 2016-06-28
Add Dockerfile alternative


sanchothefat:

rmccue:
I messed around with this a bit, but couldn't get the Dockerfile to build properly. The entirety of it is:
```
FROM ubuntu:12.04

# Install pre-requisites
RUN "./puppet/preprovision.sh"

RUN puppet apply \
	--module_paths "./puppet/modules" \
	--hiera_config /dev/null \
	--disable_warnings=deprecations \
	./puppet/manifests/development.pp
```

The problem is, the Dockerfile can't resolve `./`. Still working out how to achieve this.

rmccue:
(This is using the vagrant docker provider.)

roborourke:
Ah ok. I'll try this again but we'd probably need to set the `WORKDIR`
explicitly and then copy everything over to the container with `COPY`

rmccue:
The problem is that the directory with all the stuff in it is mounted dynamically by Vagrant at a randomised path, so we can't set WORKDIR there. Might be able to on the build call though?

rmccue:
Oh, I may be dumb; I think I need to `COPY` first.

PR #272, opened 2016-12-02
Add first pass at Docker support

rmccue:
See #193.

Not yet working. :( The problem is that everything generates upstart scripts, but phusion/baseimage uses runit. Apparently, upstart can't run inside Docker, so this is painful.

I suspect this can be resolved by upgrading to a newer base version of Ubuntu (16.04, e.g.), as it would use systemd instead of Upstart. That's a bigger topic, however.


joehoyle:
Are you thinking we'll continue using Vagrant? Not sure if that's necessary, or if that makes an additional linux VM in VB?

rmccue:
> Are you thinking we'll continue using Vagrant? Not sure if that's necessary, or if that makes an additional linux VM in VB?

Sticking with Vagrant provides a consistent CLI, and also ensures there's support for Docker even on systems without it with the VM. You can use the native Docker without the VM though if you have it installed (although that's not in this patch).

joehoyle:
Ok, I would _really_ like to not install virtualbox / vmware if at all possible.

rmccue:
See https://www.vagrantup.com/docs/docker/configuration.html#force_host_vm - Vagrant will use Docker natively without a VM if that's set to `false`

Dockerfile
```diff
@@ -0,0 +1,39 @@
+# FROM ubuntu-upstart:precise
+# http://phusion.github.io/baseimage-docker/
+FROM phusion/baseimage:0.9.9
+
+# Install pre-pre-requisites.
+RUN apt-get update
```
BronsonQuick: 🔢 Can we chain commands instead? e.g. `RUN apt-get update && apt-get install -q -y wget #ca-certificates`

Seeing I've been working with Docker on a client project I've been "levelling up" on Docker training today and the consensus I've seen is that Docker folks tend to prefer chaining as it leaves less overhead.

rmccue: This isn't suitable for merge in any case, but best practice is to merge commands where it makes sense; it doesn't with update and install (IMO). Changing the command invalidates the intermediate container, but adding a new package shouldn't invalidate the `update`, usually.

BronsonQuick: Yeah I know it's a super old PR and we'll have to update it. I came back and edited my comment from 🔢 as it only seems to make sense to chain `RUN` commands based on what Docker does behind the scenes.

I'm still doing a Docker course as we speak at the moment so I'm getting my head around all the different pieces as I'm keen to have a decent opinion about how Docker could work with Chassis and all our extensions. Long weekends FTW! I get to learn things!

rmccue: Unfortunately, it can't: systemd can't run inside a container properly.

BronsonQuick: Gotcha. When I learnt about Docker needing things to not be daemonized I figured that was probably a show stopper 😞

BronsonQuick: For peeps playing along at home I'm incorrect with that comment! @rmccue mentioned this in our Slack.

>Docker can handle daemonisation fine
It's just systemd that doesn't like not having access to the root filesystem

ntwb:
@rmccue Have you looked at https://github.com/ibuildthecloud/systemd-docker ?

cstrouse:
Any thoughts on adding LXD support? That's probably a better fit for this job than Docker.

BronsonQuick:
@cstrouse Oohh. I hadn't even heard of LXD before now. Looks like an interesting idea. I'll try and have a good read and experiment with it later this week and see if it's an option. Thanks for mentioning it!

BronsonQuick:
I think I'll close this one as I can't see us getting around to this.

PR #276, opened 2016-12-05
 Always install XML and mbstring extensions

mattheu:
I ran into issues with `mbstring` functions not being available on PHP7, which were fixed by https://github.com/Chassis/Chassis/issues/275.

However I am still experiencing issues using these functions when NOT using PHP 7.

Additionally, I also need the `xml` extension installed for all PHP versions. I have updated this to remove the mbstring extensions added earlier for PHP 7 only, and added both these for all versions of PHP.

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -3,7 +3,7 @@ import "/vagrant/extensions/*/chassis.pp"
 
 $config = sz_load_config()
 $extensions = sz_extensions('/vagrant/extensions')
-$php_extensions = [ 'curl', 'gd', 'mysql' ]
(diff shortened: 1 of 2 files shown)


rmccue:
Since xml isn't required for WP, I think it's probably better in an extension, ala memcache or xdebug.

BronsonQuick:
I'll make sure mbstring is installed in all versions of PHP. For some reason I thought it was packaged in 5.x versions! I'll tackle that here: https://github.com/Chassis/Chassis/issues/274

Once that's done I'll get to work on an XML extension.

BronsonQuick:
I've added a new Chassis extension for xml: https://github.com/Chassis/xml I'll update the docs now too.

rmccue:
Renamed back, since this is a PR. Superseded by #274 and the above extension.

PR #283, opened 2016-12-06
Allow overriding VM properties

### LINKED ISSUE #141: Manage memory allocation in YAML config
It would be great if the amount of memory used by the Vagrant could be managed in the YAML configuration file.

### LINKED ISSUE #222: Change memory and cpu
How can I change the default allocated memory and cpu ?

rmccue:
Fixes #222, fixes #141.

I attempted to do this in an extension, which quickly rabbit-holed. Turns out, it's likely not possible.


rmccue:
Now including docs changes as well, and brought up-to-date with master.

PR #293, opened 2016-12-12
Add version 2 of extensions API

rmccue:
This PR introduces a brand new extensions API that will help us deal with the lack of `import` in newer versions of Puppet (#115). It achieves this by removing the `chassis.pp` file and instead requiring extensions to have a Puppet module named the same as the extension. The init.pp class receives `config` as a property containing the Chassis configuration, which also avoids duplication that we currently have across extensions.

To mark an extension as a v2 extension, you add a `chassis.yaml` to the extension which specifies `version: 2`. In the future, we can use this YAML file to specify additional capabilities and newer versions if we need to bump it again.


docs/extend.rst
```diff
@@ -4,56 +4,20 @@ Extending Chassis
...
-* `Debugging`_: A Chassis extension to install and activate common WordPress plugins used for debugging during development.
-
-* `NodeJS`_: A Chassis extension to install NodeJS.
-
-* `Imagick`_: A Chassis extension to install ImageMagick.
+Example: Installing The Memcached Extension
+-------------------------------------------
 
-* `Fish`_: A Chassis extension to install Fish Shell.
+To install the memcached extension you would do the following:
```
BronsonQuick: I reckon it's worth either mentioning how to add extensions in the yaml files or linking to our instructions about how to do that as well as having the cloning method.

rmccue: Docs are updated in #356 which has deeper changes there; intention is to merge both into 2.0.

BronsonQuick:
Closing in favour of the updated PR: https://github.com/Chassis/Chassis/pull/398

rmccue:
@BronsonQuick Back to you for final review.

Issue #300, opened 2016-12-14
Extensions provision fails on VMWare


paulgibbs:
With VMWare, Chassis installs and works out of the box fine. I can switch between PHP versions, and it all works.

When I try to add any extension, however, it fails with the following output: 

> $ vagrant provision
==> default: Running provisioner: shell...
    default: Running: /var/folders/tr/1zvk31cs3_bds22b4gbwfdc40000gn/T/vagrant-shell20161213-3154-ad7yjt.sh
==> default: stdin: is not a tty
==> default: Running provisioner: puppet...
==> default: Running Puppet with development.pp...
==> default: stdin: is not a tty
==> default: Notice: Compiled catalog for vagrant.local in environment production in 1.33 seconds
==> default: Notice: /Stage[main]/Tester::Config/File[/vagrant/extensions/tester/wpdevel/wp-tests-config.php]/ensure: defined content as '{md5}8f838da9b4524f8eba2cdbc50c0ea819'
==> default: Error: Operation not permitted - /vagrant/local-config-extensions.php20161213-28338-ret4ba
==> default: Error: /Stage[main]/Main/Chassis::Wp[vagrant.local]/File[/vagrant/local-config-extensions.php]/content: change from {md5}d28bac1939d50ee28f050d9ce5a20713 to {md5}d2811a6204141c81798734aca9fe234d failed: Operation not permitted - /vagrant/local-config-extensions.php20161213-28338-ret4ba
==> default: Notice: Finished catalog run in 1.24 seconds
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.

After much trial and error on Slack, Ryan, Bronson, and I, discovered that if I deleted `local-config-extensions.php`  before running provision, the provision works fine and it rebuilds the deleted file.

BronsonQuick:
Thanks for the solid bug report @paulgibbs! I'll be buying VMware either this week or next week so I can get to the bottom of this and make sure we support both!= VirtualBox and VMware!

BronsonQuick:
@paulgibbs I'm going to close this one for now as I know this issue is fixed in https://github.com/Chassis/Chassis/pull/408 using VWWare and `vagrant plugin install vagrant-bindfs`

BronsonQuick:
Cancel that. I didn't test that correctly!

BronsonQuick:
Closing in favour of https://github.com/Chassis/Chassis/issues/728

PR #301, opened 2016-12-23
Add no_root_squash to synched folders

BronsonQuick:
@paulgibbs I've just installed VMWare fusion and I could replicate the error you were having in #300 . Could you please pull down this branch and test it and let me know if if fixes your issue? I'm not sure if Vagrant is smart enough to alter the folder permissions on an existing install so this might only work on a fresh up. Cheers!


paulgibbs:
Same error, I'm afraid. Tested it on the checkout I was playing with, and also on a brand-new clone.

Issue #320, opened 2017-01-23
Default hosts don't get added to /etc/chassis-hosts/conf.d/subdomains on provision


stuartshields:
So when using `multisite: subdomains`, the `hosts` from `config.yaml` don't get added to `Multisite subdomains don't get added to /etc/chassis-hosts/conf.d/subdomains`. 

Only when you go Network Admin > Sites and create a subdomain will it get added. While this might be the preferred way, when using the following code:

```
	wp::plugin { $plugins:
		location => '/vagrant/wp',
                networkwide => true,
		require  => Class['wp'],
	}
```

Will error out the following code on provision:

```
==> default: Notice: /Stage[main]/Main/Chassis::Wp[vagrant.local]/File[/vagrant/local-config-extensions.php]/content: content changed '{md5}b24989b3ec57d8f506c316d6d84f3e84' to '{md5}fcbdc6ed7f257edae64dedcbeeb82ae5'
==> default: Notice: /Stage[main]/Vip::Plugins/Wp::Plugin[keyring]/Exec[wp install plugin keyring]/returns: Error: Site vagrant.local/ not found.
==> default: Error: /usr/bin/wp plugin install keyring returned 1 instead of one of [0]
```

Take note on: Error: **Site vagrant.local/ not found.**

BronsonQuick:
Fixed. Props @stuartshields

PR #323, opened 2017-01-27
320-fix-multisite-hosts-provision - Add init hook

stuartshields:
@BronsonQuick Have a look at this implementation, I've been using it with my extension and it appears to fix everything in #320


puppet/modules/chassis/files/local-config-hosts.php
```diff
@@ -14,6 +14,17 @@
 function bootstrap() {
 	add_filter( 'wpmu_new_blog', __NAMESPACE__ . '\\add_host' );
 	add_filter( 'delete_blog', __NAMESPACE__ . '\\remove_host' );
+    add_action( 'init', __NAMESPACE__ . '\\maybe_add_host' );
```
BronsonQuick: Tabs not spaces please! https://www.youtube.com/watch?v=SsoOG6ZeyUI

stuartshields: @BronsonQuick How did I know you were going to link to that video? That is the result of me copying and pasting. Fixing it now.

PR #332, opened 2017-02-10
Skip multisite constants when installing

rmccue:
Don't merge just yet, still double-checking on this. Should fix an error with installation during running of tests.

### DIFF
--- a/puppet/modules/chassis/templates/local-config-db.php.erb
+++ b/puppet/modules/chassis/templates/local-config-db.php.erb
@@ -13,14 +13,19 @@ defined( 'ABSPATH' ) or define( 'ABSPATH', '<%= @wpdir %>/' );
 defined( 'WP_CONTENT_DIR' ) or define( 'WP_CONTENT_DIR', '<%= @contentdir %>' );
 
 <% if @network != false %>
-define( 'MULTISITE', true );
+if ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) {
+	define( 'MULTISITE', true );
 <% if @subdomains != false -%>
-define( 'SUBDOMAIN_INSTALL', true );
+	define( 'SUBDOMAIN_INSTALL', true );
 <% else -%>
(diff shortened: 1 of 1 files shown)


rmccue:
@BronsonQuick Can you give this a 👀 and see if it seems sensible to you? I can't think of any way it would, but testing now to make sure.

puppet/modules/chassis/templates/local-config-db.php.erb
```diff
@@ -13,14 +13,19 @@ defined( 'ABSPATH' ) or define( 'ABSPATH', '<%= @wpdir %>/' );
 defined( 'WP_CONTENT_DIR' ) or define( 'WP_CONTENT_DIR', '<%= @contentdir %>' );
 
 <% if @network != false %>
-define( 'MULTISITE', true );
+if ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) {
+	define( 'MULTISITE', true );
```
BronsonQuick: Shouldn't this be wrapped in a conditional? e.g. `<% if subdomains != "No" %>`

puppet/modules/chassis/templates/local-config-db.php.erb
```diff
@@ -13,14 +13,19 @@ defined( 'ABSPATH' ) or define( 'ABSPATH', '<%= @wpdir %>/' );
...
 <% if @subdomains != false -%>
```
BronsonQuick: I just noticed there is a stray `-` before the closing tag.

rmccue: This is correct: it swallows the newline after the tag to avoid outputting too much whitespace

puppet/modules/chassis/templates/local-config-db.php.erb
rmccue: It's already inside `<% if @network != false %>` just above, which is the internal name for the `multisite` config option?

BronsonQuick:
Coolio. Merged!

PR #315, opened 2017-01-15
Bump WP to 4.7.1

peterwilsoncc:
Upgrade WP to 4.7.1

PR #339, opened 2017-03-07
Update core to 4.7.3

cameronjonesweb:
New security and maintenance release this morning https://wordpress.org/news/2017/03/wordpress-4-7-3-security-and-maintenance-release/


BronsonQuick:
@cameronjonesweb Thanks heaps for the PR! Before I merge it I just wanted to check how you updated this? Did you update the subtree?
e.g. `git subtree pull --prefix=wp --squash git@github.com:WordPress/WordPress 4.7.3` or did you just click update in the backend?

If it's the former then we're good to merge if it's the latter then it'll break the reference to the subtree. I got caught out by that before when I merged in https://github.com/Chassis/Chassis/pull/315 and it broke Chassis 😱

cameronjonesweb:
Didn't update the subtree :( didn't know that was a thing. 
Though when I rollback and try to update the subtree it says `Working tree has modifications. Cannot add` :(

BronsonQuick:
>Didn't update the subtree :( didn't know that was a thing.
Yeah sorry. It's a bit of a weird one. Not many people use subtrees in Git repos. It decreases the provisioning time which is why we do it.

>Working tree has modifications. Cannot add
I'll take care of it for you. Hang 10 😄  Thanks again for the PR. Most appreciated!

BronsonQuick:
@cameronjonesweb I've pushed up a new branch and I'm just testing a fresh provision. For future reference seeing subtrees aren't used much. To update WordPress you need to delete the `wp` folder then commit that change then once that's committed you can run `git subtree add --prefix=wp --squash git@github.com:WordPress/WordPress.git 4.7.3` then push that up. I'll close this PR for now and have a new PR in a tick. Once again thanks again for being proactive! 😍

PR #341, opened 2017-03-07
Include the {fqdn}.d directory in multisite's Nginx config

johnbillion:
The [OpenSSL module for Chassis](https://github.com/javorszky/chassis-openssl) doesn't work with Multisite enabled. The cause is that the `{fqdn}.d` directory is not included in the Nginx config for multisite [like it is for single sites](https://github.com/Chassis/Chassis/blob/8186f25bce9ff40345c86811c063a575c747e1c8/puppet/modules/chassis/templates/site.nginx.conf.erb#L11), therefore the host name isn't made available on port 443.

Adding this line to the Nginx config for Multisite means the OpenSSL module works as expected.

### DIFF
--- a/puppet/modules/chassis/templates/multisite.nginx.conf.erb
+++ b/puppet/modules/chassis/templates/multisite.nginx.conf.erb
@@ -7,6 +7,8 @@ server {
 	fastcgi_split_path_info ^(.+\.php)(/.+)$;
 	fastcgi_index index.php;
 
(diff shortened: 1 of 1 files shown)


BronsonQuick:
@rmccue I had a brief chat to @johnbillion about this and I think this is fine but I did tell say you're more likely to know if this could cause an issue. I'll leave this one to you do review please. 🙇

PR #342, opened 2017-03-07
Add extensions to the config docs

johnbillion:


BronsonQuick:
@johnbillion Thank you Sir! I didn't realise we hadn't documented that 😱 I knew we'd listed the extensions [here](http://docs.chassis.io/en/latest/extend/) but wasn't aware we hadn't documented how easily it is to have them installed in our yaml files.

Keep the PR's coming through 😍. It's helping me realise how much assumed knowledge I have with Chassis.

PR #345, opened 2017-03-16
Use correct site URL for multisite

roborourke:
Currently a multisite install is set up with the wrong `WP_SITEURL` constant. The result is that although you can access the multisite at the root URL eg. `http://vagrant.local` all the generated links have `/wp` appended to them.


rmccue:
Closing this in favour of #351 instead.

PR #348, opened 2017-03-17
Support minor versions correctly

### LINKED ISSUE #346: Unable to use PHP version 7.0.16
I'm just starting to get started with Chassis, but ran into this small issue when I tried to change my configuration to match a user's reported issue with the BWP plugin. This specific version 7.0.16 was the one supposedly having issues, but when I tried to update my local config file it fails to find the correct file (I think?): 

```Error: /Stage[main]/Chassis::Php/Package[php7.0.16-common]/ensure: change from purged to latest failed: Could not update: Execution of '/usr/bin/apt-get -q -y -o DPkg::Options::=--force-confold --force-yes install php7.0.16-common' returned 100: Reading package lists...
==> default: Building dependency tree...
==> default: Reading state information...
==> default: E: Unable to locate package php7.0.16-common
==> default: E: Couldn't find any package by regex 'php7.0.16-common'
==> default: Warning: /Package[php7.0.16-cli]: Skipping because of failed dependencies```

Do you need any other information? I switched back to PHP version 7.0

rmccue:
Fixes #346.

### DIFF
--- a/puppet/modules/chassis/manifests/php.pp
+++ b/puppet/modules/chassis/manifests/php.pp
@@ -7,9 +7,11 @@ class chassis::php (
 
 	if $version =~ /^(\d+)\.(\d+)$/ {
 		$package_version = "${version}.*"
+		$short_ver = $version
 	}
 	else {
 		$package_version = "${version}*"
+		$short_ver = regsubst($version, '^(\d+\.\d+)\.\d+$', '\1')
 	}
 
 	if versioncmp( "${version}", '5.4') <= 0 {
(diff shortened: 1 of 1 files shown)


BronsonQuick:
I'm going to test this with a few random php versions and check them off as I go:
- [ ] 7.0.16
- [ ] 5.6
- [ ] 7.1.3
- [ ] 5.6.30
- [ ] 7.0

BronsonQuick:
I've tested a few versions and this is good to merge. If the version isn't in the ppa the closest version to that will be installed. In @katmoody's case in #346 She will get 7.0.17

PR #356, opened 2017-03-28
Add dependencies for extensions.

bradp:
This probably needs more testing, and I'm not _great_ with Ruby, so there could probably be some cleanup that could be done there. 

This PR adds the ability to define dependencies for extensions in a bit easier way (two different ways), as well as a new pattern for putting configs from an extension that may need to be processed in a different way than just the puppet module.

First, the easy one, the config file dependency. [Some](https://github.com/Chassis/Gulp) [of](https://github.com/Chassis/Yarn) [our](https://github.com/Chassis/Grunt) [extensions](https://github.com/Chassis/Bower) have dependencies, and right now the process 'make sure you also list the dependency in your config. This PR adds this format for defining dependencies, so its a bit clearer (to me at least):
```
# <my-chassis-clone>/config.yaml or config.local.yaml
# 
# Extensions
extensions:
  - gulp: ['nodejs']
```
(This PR also makes it so that if your extension is just a single word, we assume its from the Chassis GH org, so you can skip the `chassis/nodejs` and just do `nodejs`. This makes it a bit more readable I think.

The second way of doing dependencies I'm a bit more excited about. If you have a `chassis.yaml` file in the root of your extension, that file will get loaded and processed. An example of this for the Grunt extension might look like this:
```
# <my-extension>/chassis.yaml
# 
# Dependencies
dependencies:
  - 'nodejs'
```

Chassis will now read that and pull in the dependencies if it needs them. By keeping the dependencies in the actual extension, we no longer need to tell the user what the dependencies are and to make sure they install them. This could also let you have project-specific extensions that pull in all the required extensions for that project, without having to copy those into your config.

In the future, we could also expand on this `chassis.yaml` file to have things other than dependencies.

Last note - the docs I wrote up while I was working on this, as somewhat of a note-to-myself, so those probably will need some edits before they're 100% good to go.


puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
+
+		# For each of the extensions in our folder, read the extension config and
+		# install dependencies, etc for that extension.
+		Dir.foreach('extensions/') do |ext_folder|
+
+			# Bail out if we aren't a valid folder.
+			next if ext_folder == '.' or ext_folder == '..' or File.file?('extensions/' + ext_folder)
+
+			# If we don't have our config file, don't do anything.
+			next if not File.exists? 'extensions/' + ext_folder + '/chassis.yaml'
```
BronsonQuick: 🔢 Rather than `chassis.yaml` let's go with `config.yaml` instead. I've tested using that filename in a branch and it works fine! 🎉

BronsonQuick:
@bradp I ❤️ this PR! Great work!

I've tested both configs and can confirm they work for me. I left one comment about a change I'd like you to make but it's only minor.

I'll add @rmccue as a reviewer to get his thoughts on this PR but it's a ➕ from me.

rmccue:
tl;dr: 🛑 Hold on! 🛑

I love the idea here, and think this could work well, *but* note that #293 *also* adds an `{extension}/chassis.yaml`. These two need to work together.

Also, 👎 on the main config file format: I don't think it's necessary if we add the extension config, and there should be no need for end-users to worry about it.

Will leave some inline comments on the actual code too!

puppet/chassis.rb
rmccue: 👎 per other comment.

puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
-			config["extensions"].each do |ext|
-				begin
-					repo = /^(git@|https:)/.match(ext) ? ext : 'https://github.com/' + ext
-					folder = @@dir + '/extensions/' + ext.split('/').last.gsub(/.git$/, '')
+			self.install_extensions_from_array(config["extensions"])
+		end
+
+		# For each of the extensions in our folder, read the extension config and
+		# install dependencies, etc for that extension.
+		Dir.foreach('extensions/') do |ext_folder|
```
rmccue: We should use `get_extensions` from #293.

puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
+
+			# If we don't have our config file, don't do anything.
+			next if not File.exists? 'extensions/' + ext_folder + '/chassis.yaml'
+
+			# Grab our config file and load it.
+			ext_config = YAML.load_file('extensions/' + ext_folder + '/chassis.yaml')
+
+			# If we have dependencies, then install them.
+			if ext_config["dependencies"]
+				install_extensions_from_array(ext_config["dependencies"])
```
rmccue: Should be `install .... if ext_config[...]` in one line

puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
+				ext.keys.each do |ext_need_dep|
+					# Loop through each of the dependencies we need
+					# for the extension and install them.
+					ext.values[0].each do |ext_dep|
+						self.install_extension(ext_dep)
 					end
+
+					# Finally, actually install our extension
+					# that needs dependencies.
+					self.install_extension(ext_need_dep)
```
rmccue: Per other comment, should remove these bits.

puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
+		# Perform checks for various forms of extension definition.
+		if extension.include? "git@" # Full git-based url.
+			repo = extension
+		elsif extension.include? 'https:' # GitHub-like https url.
+			repo = extension
+		elsif extension.include? '/' # assume account/repo style.
+			repo = 'https://github.com/' + extension
+		else # Assume a Chassis official extension, like 'nodejs'.
+			repo = 'https://github.com/chassis/' + extension
+		end
```
rmccue: I'd prefer swapping this check: if it's `xxxx` or `xxxx/xxxx`, assume GitHub, all else is a URL. (Where `x` is `/[\w-]/`)

puppet/chassis.rb
```diff
@@ -102,16 +102,68 @@ def self.config
...
+
+		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '')
+
+		if ! File.exist?( folder )
+			system("git clone %s %s --recursive" % [repo, folder] )
```
rmccue: `system( ... ) unless File.exist?( folder )`

rmccue:
@bradp I've updated your PR with the changes I suggested above, and based it off the new extensions branch. Wanna give it a 👁?

docs/extend.rst
```diff
@@ -90,6 +90,11 @@ Extensions **must** specify the following options:
 
 - ``version``: Extension API version. Currently ``2``.
 
+Extensions may also specify the following options:
+
+- ``dependencies``: Extensions that the extension depends on. List of strings,
+  where each string is an extension name (see :ref:`extension-format-ref`).
```
rmccue: Fingers crossed this works! We'll find out after merge. :) (I can't enable docs builds for an external PR.)

bradp:
@rmccue Just updated with some fixes for deps inside of the main config file, other than that it's looking perfect!

puppet/chassis.rb
```diff
@@ -129,16 +129,39 @@ def self.config
...
+		elsif extension =~ /^[\w-]+\/[\w-]+$/ # account/repo style.
+			repo = 'https://github.com/' + extension
+		else
+			repo = extension
+		end
+
+		if extension.is_a?(String)
+			folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '')
+			system("git clone %s %s --recursive" % [repo, folder] ) unless File.exist?( folder )
+		elsif extension.is_a?(Hash)
```
rmccue: We assume `extension` is a string in the above checks, so this feels like too late to check. What does this fix specifically that isn't already handled? 🤔

bradp: It was to allow for deps to be added in-line when defining them.
```
# <my-chassis-clone>/config.yaml or config.local.yaml
# 
# Extensions
extensions:
  - gulp: ['nodejs']
```

But now that I look at it again, it seems kind of pointless and dumb, you can just add it to the normal list.

BronsonQuick:
Once https://github.com/Chassis/Chassis/pull/398 is merged I'll have to get you to do another PR for this @bradp into `master`. We shipped Xenial faster than we expected seeing the ppa's "broke" Chassis for a few days.

rmccue:
@bradp Per above comments, reverting ee6463796e8b233a3ad3228a434504809bd4992d. This is already handled at a higher level where needed, and I don't want to allow dependencies in `config.yaml` :)

PR #357, opened 2017-03-28
Remove example extension folder

bradp:
As we have a ton of fully-built extensions, I don't know if we need to keep the empty sample extension folder around anymore


rmccue:
This is partially intentional to ensure the extensions dir always exists, and you have one in there all the time.

bradp:
Makes sense.

PR #358, opened 2017-03-28
Clone extensions shallowly

bradp:
Right now, most extensions are pretty small, so doing a full git clone isn't a huge deal, but we might as well just clone them shallowly, as it is a bit faster.


bradp:
Something is causing a few errors locally when sshing in, not 100% sure if it is due to this change or something else. Going to close this to investigate that, will reopen.

ntwb:
This sounds good to me, testing this with https://github.com/Chassis/Tester would probably be the ultimate test to see if the WP Git sub-module breaks using this, hopefully not 😏

Issue #360, opened 2017-03-29
Restart needed when changing default admin


LewisCraik:
When changing default admin user/password a full destroy/up is needed, rather than just a provision. This should be included in the documentation.

PR #361, opened 2017-03-29
Update documentation for changing default admin

LewisCraik:
Addressing issue #360 - added line to documentation indicating that vagrant destroy then vagrant up is needed when changing default admin details.


BronsonQuick:
@LewisCraik Good catch. Thanks so much for the PR! ✨

PR #363, opened 2017-04-09
remove wordpress and add shell script

micfunk:


micfunk:
Sorry about that. Opened on accident.

Issue #364, opened 2017-04-10
Branch a 2.0 for development


rmccue:
Let's start a 2.0 development branch (`two-dot-oh`). Things to merge essentially straight away:

* [x] Extensions v2: #293 
* [x] Xenial: #312

rmccue:
Branched, ready for merging.

BronsonQuick:
@rmccue Looks like it needs a PR and a push? https://github.com/Chassis/Chassis/tree/two-dot-oh

PR #370, opened 2017-04-26
Merge 2.0

rmccue:
From #364.

Don't merge yet, this is just to track our overall changes.


rmccue:
FYI, I've also turned on docs builds for this branch so we can preview those: http://docs.chassis.io/en/two-dot-oh/

PR #371, opened 2017-05-04
Added nfs to Vagrant config

mikeselander:
Added NFS for the performance boost that can be received when running PHPUnit tests from within the machine. Thanks @BronsonQuick for helping me with the fix and the formatting of the option.

Only added it to the `> 1.3.0` config due to possibility of issues on old installs of Windows machines.

Reference: https://www.vagrantup.com/docs/synced-folders/nfs.html


rmccue:
IMO, this should probably be behind a configuration flag, since running NFS requires sudo on the host, and is generally a worse experience. Most sites (that aren't filesystem-access heavy) probably don't need NFS.

PR #376, opened 2017-05-14
Revert "Added nfs to Vagrant config"

### LINKED ISSUE #375: NFS fails on vagrant up
## Expected
Provisioning a new VM should map my local Chassis directory to `/vagrant` on the VM.

## Actual
Provisioning succeeds with errors; NFS fails to map Chassis to `/vagrant` and logs to `/var/log`.

## Summary
On a new install of Chassis, vagrant throws an error and fails to fully map to `/vagrant`, so I'm left with a running VM with an empty `/vagrant` directory and no WordPress install.

Here's the error messages:

```sh
==> default: Exporting NFS shared folders...
==> default: Preparing to edit /etc/exports. Administrator privileges will be required...
==> default: Mounting NFS shared folders...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mount -o vers=3,udp 172.28.128.1:/Users/mestes/Sites/obyp-chassis/Chassis /vagrant
result=$?
if test $result -eq 0; then
if test -x /sbin/initctl && command -v /sbin/init && /sbin/init 2>/dev/null --version | grep upstart; then
/sbin/initctl emit --no-wait vagrant-mounted MOUNTPOINT=/vagrant
fi
else
exit $result
fi


Stdout from the command:



Stderr from the command:

mount.nfs: requested NFS version or transport protocol is not supported
```

I've verified that `/etc/exports` on my host machine has the expected entries, the `/vagrant` directory exists on the VM (empty), and `nfsd` is running on host (macOS Sierra 10.12.4).

Reverting changes made in #371 allows me to provision the VM and get a working dev environment.

BronsonQuick:
Reverts Chassis/Chassis#371

Let's revert this for now because of https://github.com/Chassis/Chassis/issues/375. Let's readd it behind a config flag as per Ryan's suggestion here: https://github.com/Chassis/Chassis/pull/371#issuecomment-299752228

Fixes #375

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -111,9 +111,9 @@ Vagrant.configure("2") do |config|
 
 	# Ensure that WordPress can install/update plugins, themes and core
 	if vagrant_version >= "1.3.0"
-		config.vm.synced_folder ".", "/vagrant", :nfs => { :mount_options => ["dmode=777","fmode=777"] }
+		config.vm.synced_folder ".", "/vagrant", :mount_options => [ "dmode=777,fmode=777" ]
 		CONF["synced_folders"].each do |from, to|
-			config.vm.synced_folder from, to, :nfs => { :mount_options => ["dmode=777","fmode=777"] }
+			config.vm.synced_folder from, to, :mount_options => [ "dmode=777,fmode=777" ]
 		end if CONF["synced_folders"]
 	else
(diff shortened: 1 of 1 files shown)


BronsonQuick:
Fixes #375

PR #381, opened 2017-05-22
Add NFS behind a flag

### LINKED ISSUE #377: Add support for NFS behind a config flag
We need to add back https://github.com/Chassis/Chassis/pull/371 but make sure it's behind a config flag instead.

BronsonQuick:
Fixes #377

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -117,10 +117,17 @@ Vagrant.configure("2") do |config|
 
 	# Ensure that WordPress can install/update plugins, themes and core
 	if vagrant_version >= "1.3.0"
-		config.vm.synced_folder ".", "/vagrant", :mount_options => [ "dmode=777,fmode=777" ]
-		CONF["synced_folders"].each do |from, to|
-			config.vm.synced_folder from, to, :mount_options => [ "dmode=777,fmode=777" ]
-		end if CONF["synced_folders"]
+		if CONF['virtualbox']['nfs']
+			config.vm.synced_folder ".", "/vagrant", :nfs => { :mount_options => ["dmode=777","fmode=777"] }
+			CONF["synced_folders"].each do |from, to|
+				config.vm.synced_folder from, to, :nfs => { :mount_options => ["dmode=777","fmode=777"] }
+			end if CONF["synced_folders"]
+		else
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -117,10 +117,17 @@ Vagrant.configure("2") do |config|
 
 	# Ensure that WordPress can install/update plugins, themes and core
 	if vagrant_version >= "1.3.0"
-		config.vm.synced_folder ".", "/vagrant", :mount_options => [ "dmode=777,fmode=777" ]
-		CONF["synced_folders"].each do |from, to|
-			config.vm.synced_folder from, to, :mount_options => [ "dmode=777,fmode=777" ]
-		end if CONF["synced_folders"]
+		if CONF['virtualbox']['nfs']
```
rmccue: Does this apply just to VB, or can we use it for VMware as well?

BronsonQuick: I was wondering about that when I coded it up. I decided on VB only because @joehoyle mentioned in Slack he was having issues with NFS on VMWare but I didn't have any specific details.

@joehoyle What are your thoughts Sir? VB only or should this flag enable it for both?

joehoyle: I'd flag it for both. For what it's worth, though I'm using VMware now I used to use VirtualBox a lot in the past and always ran into nasty caching issues with NFS - it's possible that doesn't exist now, but I believe it's something inherent with the OS X file watching with NFS that causes this, not an issue in VMWare.

rmccue: AFAIK, the `nfs` flag isn't specific to VB anyway, so we should let the user decide whether they want to deal with NFS or not. :)

Vagrantfile
```diff
@@ -117,10 +117,17 @@ Vagrant.configure("2") do |config|
...
+			config.vm.synced_folder ".", "/vagrant", :nfs => { :mount_options => ["dmode=777","fmode=777"] }
+			CONF["synced_folders"].each do |from, to|
+				config.vm.synced_folder from, to, :nfs => { :mount_options => ["dmode=777","fmode=777"] }
```
rmccue: Is there no better way to do this? Surprised that there's nothing like a `:type` argument rather than setting it this way.

rmccue:
Closing in favour of #406.

PR #383, opened 2017-05-28
Show php all errors in WP_CLI

BronsonQuick:
When we're running WP CLI commands it's handy to show all php error logs rather then the default which suppresses those errors.

### DIFF
--- a/local-config-sample.php
+++ b/local-config-sample.php
@@ -10,3 +10,8 @@ define( 'AUTOMATIC_UPDATER_DISABLED', true );
 
 // You'll probably want debug logging during development
 define( 'WP_DEBUG_LOG', true );
+
+// Show all php errors for WP_CLI
+if ( defined( 'WP_CLI' ) && WP_CLI ) {
+	ini_set( 'display_errors', 1 );
+}


rmccue:
Can we handle this at the php.ini level instead? See https://github.com/Chassis/Chassis/pull/374#pullrequestreview-38825173

BronsonQuick:
I considered doing that but the reason I didn't and went with this approach was that I was concerned logging all PHP errors, warnings and notices could cause massive error logs. I've been shocked by the number of PHP SDK's and libraries that throw notices so I thought only having this for WP CLI would be better.

Let me know what you think as I'm happy to go either way and I'll trust your judgement!

rmccue:
> I considered doing that but the reason I didn't and went with this approach was that I was concerned logging all PHP errors, warnings and notices could cause massive error logs.

Those are already getting logged per the change on #374, so not sure how that's related?

BronsonQuick:
Haha valid point is valid. #derp I'll rework this later in the week! Thanks!

BronsonQuick:
We can close this as https://github.com/Chassis/Chassis/pull/374 was merged.

PR #398, opened 2017-06-21
Extensions v2

BronsonQuick:
https://github.com/Chassis/Chassis/pull/293 is out of date now as we've merged Xenial into `master`. I'll copy across the comments that @rmccue left on that original PR as they are still relevant.

```
This PR introduces a brand new extensions API that will help us deal with the lack of import in newer versions of Puppet (#115). It achieves this by removing the chassis.pp file and instead requiring extensions to have a Puppet module named the same as the extension. The init.pp class receives config as a property containing the Chassis configuration, which also avoids duplication that we currently have across extensions.

To mark an extension as a v2 extension, you add a chassis.yaml to the extension which specifies version: 2. In the future, we can use this YAML file to specify additional capabilities and newer versions if we need to bump it again.
```


ntwb:
Excellent, the _example_ plugin and docs are great for Puppet noobs 👍

rmccue:
@BronsonQuick FWIW, it's better to edit the PR and change the base rather than opening a new PR.

rmccue:
I'm going to close this PR and change the other one over instead, because it doesn't make sense for me to review my own commits. :)

PR #402, opened 2017-06-30
Delete .htaccess

grappler:
This is a nginx system so .htaccess does not work. This seems to be remains from the initial commit from WordPress Skeleton


rmccue:
Never noticed this before, thanks!

PR #406, opened 2017-07-04
Add NFS support

BronsonQuick:
This fixes #391, fixes #392 and fixes #381! 🎉 

@joehoyle would you mind testing this for me? You'll need to add`nfs: Yes` to one of your yaml files.

PR #307, opened 2017-01-01
VMware permissions fix

BronsonQuick:
Fixes #300.

VMWare has a known issue with file permissions and provisioning. I've tried all the options we can pass into `config.vm.synced_folder` and I also tried fixing this in Puppet as well but it wouldn't work. Other people have "solved" this problem by using this Vagrant plugin: https://github.com/gael-ian/vagrant-bindfs in the spirit of keeping Chassis as lean as possible this is the best fix for this issue.

@paulgibbs Can you please pull this one down and test it for me? I'm pretty damn sure I've nailed this now after 4 hours of trying different things! Let me know how you go and if it's good I'll get Ryan to review it. Thanks!

PR #408, opened 2017-07-06
Simplify and fix the NFS configuration

rmccue:
PR on #406.

* Fixes the mount options to actually be used.
* Refactors to remove duplication.
* Automatically uses `vagrant-bindfs` per #307 if available.
* Adds documentation for the NFS setting.

Tested on:
* [x] VirtualBox with default
* [x] VirtualBox with default + vagrant-bindfs
* [x] VirtualBox with nfs
* [x] VirtualBox with nfs + vagrant-bindfs
* [ ] VMware with default
* [ ] VMware with default + vagrant-bindfs
* [x] VMware with nfs
* [x] VMware with nfs + vagrant-bindfs

To test: vary `nfs` option in YAML, and try with the bindfs plugin installed (`vagrant plugin install vagrant-bindfs` `vagrant plugin uninstall vagrant-bindfs`).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -106,24 +106,21 @@ Vagrant.configure("2") do |config|
 		]
 	end
 
+	# Set up synced folders.
+	synced_folders = CONF["synced_folders"].clone
+	synced_folders["."] = "/vagrant"
+
 	# Ensure that WordPress can install/update plugins, themes and core
 	mount_opts = CONF['nfs'] ? [] : ["dmode=777","fmode=777"]
 
-	config.vm.synced_folder ".", "/vagrant", :mount_options => mount_opts, :nfs => CONF['nfs']
-
-	# Automatically use bindfs if we can.
-	if CONF['nfs'] && Vagrant.has_plugin?("vagrant-bindfs")
-		config.bindfs.bind_folder "/vagrant", "/vagrant"
-	end
-
-	CONF["synced_folders"].each do |from, to|
+	synced_folders.each do |from, to|
 		config.vm.synced_folder from, to, :mount_options => mount_opts, :nfs => CONF['nfs']
 
 		# Automatically use bindfs if we can.
 		if CONF['nfs'] && Vagrant.has_plugin?("vagrant-bindfs")
 			config.bindfs.bind_folder to, to
 		end
-	end if CONF["synced_folders"]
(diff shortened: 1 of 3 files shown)


rmccue:
@BronsonQuick Can you a) review/merge, and b) test on VMware? Should only take a few minutes to test (only need to toggle options and `vagrant reload`, no need to destroy & up).

BronsonQuick:
@rmccue I just did a quick run with VMWare with default and it's got the same issue it had before which is:
```
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

mount | grep " /mnt/vagrant-mounts/1000-1000 "
if test $? -ne 0; then
mkdir -p '/mnt/vagrant-mounts/1000-1000'
chmod 700 '/mnt/vagrant-mounts'
vmhgfs-fuse -o allow_other,default_permissions,uid=1000,gid=1000,dmode=777,fmode=777 .host:/ '/mnt/vagrant-mounts/1000-1000'
fi
Stdout from the command:

Stderr from the command:

fuse: unknown option `dmode=777'
```
My other branch that you closed fixed that issue using NFS for VMWare.

I'll try and test the other variations tonight for you.

BronsonQuick:
`VMware with default + vagrant-bindfs` works a treat.

BronsonQuick:
Alrighty so after some quick testing while I was updating some other software these two aren't working:
- VMware with default
- VMware with default + vagrant-bindfs

I thinking maybe it'd be easier to throw a notice if VMWare is being used and get them to switch nfs on or we automagically flip that.

I had a bit of a Google last night and I could detect `--provider=vmware_fusion` with `ARGV` but obviously if someone only has VMWare then a `vagrant up` would use VMWare by default so we wouldn't be able to detect that argument.

rmccue:
> these two aren't working

Are those working on `master` currently? I didn't change anything for non-NFS, so it seems like these have just always been broken...

BronsonQuick:
Yeah they don't work in `master`. Chassis used to work in VMWare but that was pre-v2 I believe.

rmccue:
> Yeah they don't work in `master`.

Filed #409 to handle that since it's an existing issue; PR should be good to go, since it's not breaking anything that isn't already broken. :)

PR #410, opened 2017-07-11
Use provider-specific mount options

### LINKED ISSUE #409: VMWare default file shares are not working
@BronsonQuick noticed during testing for #408 that file shares for VMWare aren't working, due to the mount options we use. After #408 and #406 are merged, VMWare users will be able to use NFS, but otherwise it's broken.

rmccue:
Rather than assuming a one-size-fits-all approach to the mount options, let's use provider-specific ones when we need to in the overrides.

Fixes #409.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -124,13 +124,12 @@ Vagrant.configure("2") do |config|
 		end
 	}
 
-	# Each provider needs different default mount options
+	# Run with default settings first...
+	configure_synced_folders.call config, []
+
+	# ...then override for specific providers with custom mount options.
 	config.vm.provider "virtualbox" do |v, override|
 		configure_synced_folders.call override, ["dmode=777","fmode=777"]
-	end
-
-	config.vm.provider "vmware_fusion" do |v, override|
-		configure_synced_folders.call override, []
 	end
 
 	# Success?
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -107,19 +107,30 @@ Vagrant.configure("2") do |config|
 	end
 
 	# Set up synced folders.
-	synced_folders = CONF["synced_folders"].clone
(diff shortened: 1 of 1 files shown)


BronsonQuick:
This works on VirtualBox but it fails on `nginx` for VMWare. I've tried it twice to make sure it can be replicated and it can. Here's a [gist of the output](https://gist.github.com/BronsonQuick/0d59508c8bc13ef21aa65c549c965643).

rmccue:
Ah, failing on `==> default: chown: changing ownership of '/var/log/nginx': Operation not permitted`

We should probably remove the chowning there, since it won't work without bindfs.

BronsonQuick:
Interesting. I can't replicate this locally. I get this:
```
==> default: Mounting shared folders...
    default: /vagrant => /Volumes/Sites/wordcamp
    default: /var/log/php => /Volumes/Sites/wordcamp/logs/php
==> default: Detected mount owner ID within mount options. (uid: 1000 guestpath: /var/log/php)
==> default: Detected mount group ID within mount options. (gid: 1000 guestpath: /var/log/php)
    default: /var/log/nginx => /Volumes/Sites/wordcamp/logs/nginx
==> default: Detected mount owner ID within mount options. (uid: 1000 guestpath: /var/log/nginx)
==> default: Detected mount group ID within mount options. (gid: 1000 guestpath: /var/log/nginx)
    default: /tmp/vagrant-puppet/manifests-846018e2aa141a5eb79a64b4015fc5f3 => /Volumes/Sites/wordcamp/puppet/manifests
==> default: Running provisioner: shell...
    default: Running: /var/folders/8v/yvlr3qld49lcc1h837y_06gm0000gn/T/vagrant-shell20170717-5922-vht6tf.sh
```
I don't have `bindfs` installed either.

PR #420, opened 2017-07-19
Fix versioncmp issue

BronsonQuick:
Fixes #419

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ install:
   - gem install puppet-lint
 
 script:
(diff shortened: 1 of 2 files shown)


rmccue:
We need to disable the lint for this: http://puppet-lint.com/checks/only_variable_string/

BronsonQuick:
@rmccue That check caught a bunch of other cases where we were doing that which I corrected. The `versioncmp`is the only exception to the rule seeing it needs to be a string. Could we maybe cast it so that we can keep that check?

rmccue:
> Could we maybe cast it so that we can keep that check?

The string bit is how you cast in Puppet.

We could alternatively do it on the Ruby parsing side in `chassis.rb`.

rmccue:
FWIW, I actually think that sniff is dangerous, because there's no other way to cast to a string in Puppet, so it's blocking valuable functionality.

BronsonQuick:
@rmccue Updated! I guess we should tag this release too yeah?

BronsonQuick:
Fixes in #421

PR #433, opened 2017-08-13
Travis CI: Initial pass at testing Chassis install on Travis CI

ntwb:
Testing a basic install of Chassis on Travis CI

** DO NOT MERGE **

PR #434, opened 2017-08-13
Testing Chassis install on Travis CI

BronsonQuick:
** DO NOT MERGE **

I'm having a play around too based on @ntwb's PR https://github.com/Chassis/Chassis/pull/433


rmccue:
FYI, running inside Docker won't work for the same reasons as #272.

PR #441, opened 2017-08-25
Add the ${fqdn}.d directory in core

### LINKED ISSUE #440: sites-available/${fqdn}.d directory should be created in core
1. What operating system do you use?
OSX

2. What version of Vagrant are you running?
1.9.7

3. Are you using VirtualBox or VMWare and which version are you using?
VMWare

4. Do you have a custom YAML file? If so, what does it contain?
```yaml
hosts:
  - vagrant.local
  - test.local
```

5. Do you have any Chassis extensions installed? If so, what are they?
chassis-openssl

The problem is that the core nginx site config contains the include directive to include all files in `/etc/nginx/sites-available/${fqdn}.d/`. This is a problem because the `chassis-openssl` extension is the one that actually creates the directory - that should be handled by core otherwise other extensions will collide by trying to create / ensure the same directory. Even when aliased it causes provisioning to fail.

roborourke:
This fixes #440 so multiple extensions can add files to it.

### DIFF
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -21,7 +21,7 @@ define chassis::site (
 		content => template('chassis/site.nginx.conf.erb'),
 		notify  => Service['nginx']
 	}
-	file { "/etc/nginx/sites-available/${name}.d": 
+	file { "/etc/nginx/sites-available/${name}.d":
 		ensure  => directory,
 		require => Package['nginx']
 	}
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -21,9 +21,11 @@ define chassis::site (
 		content => template('chassis/site.nginx.conf.erb'),
 		notify  => Service['nginx']
 	}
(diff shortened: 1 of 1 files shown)


BronsonQuick:
@roborourke Nice catch. I've tested this and I'm happy to merge this once you remove the trailing space on the end of line 24 so our Puppet linting passes in Travis. Once we've merged this I'll leave a comment on javorszky/chassis-openssl#12 as well. Thanks for the PR! 💖

puppet/modules/chassis/manifests/site.pp
```diff
@@ -21,6 +21,10 @@
 		content => template('chassis/site.nginx.conf.erb'),
 		notify  => Service['nginx']
 	}
+	file { "/etc/nginx/sites-available/${name}.d":
```
rmccue: As a temporary fix for compatibility, can we wrap this in a `if ( ! defined( File["/etc/nginx/sites-available/${name}.d"] ) ) {` check?

roborourke: can use that in my extension too

roborourke:
@rmccue @BronsonQuick updated, cheers for reviewing :)

PR #446, opened 2017-09-02
Add a conditional so extensions can override MySQL if they need to

### LINKED ISSUE #445: We need to add the ability to override mysql::server in extensions
This is related to the https://github.com/Chassis/MariaDB extensions I made this afternoon.

BronsonQuick:
Fixes #445

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -26,8 +26,10 @@ package { 'git-core':
 	ensure => installed
 }
 
-class { 'mysql::server':
-	root_password => 'password',
+if ! defined ( Class['::mysql::server'] ) {
+	class { 'mysql::server':
+		root_password => 'password',
+	}
 }
 
 class { 'chassis':


rmccue:
This shouldn't be required. I think there are two ways to do this already; amending attributes or setting on the params class.

[Amending attributes](https://docs.puppet.com/puppet/5.1/lang_resources_advanced.html#amending-attributes-with-a-resource-reference):

```puppet
Class['mysql::server'] {
    package_name => 'mariadb-server',
    # ...
}
```

(Should be able to override everything except `root_password`)

Alternatively, since we don't declare it, you should be able to declare the `mysql::params` class:

```puppet
class mysql::params {
    server_package_name => 'mariadb-server',
    # ...
}
```

BronsonQuick:
@rmccue Thanks for the review. I'll see if I can rework my extension based on your feedback!

rmccue:
@BronsonQuick What's the status of this one? Can we close?

BronsonQuick:
I couldn't get it to work so yeah I'll kill it!

Issue #390, opened 2017-06-07
Build a base box for Chassis


joehoyle:
I've often complained to @rmccue and @BronsonQuick about what I think is pretty unacceptable download and provision times for chassis, but after yet another time of having to set up chassis I think it's important that this be detailed properly.

Given chassis by design is "one per project" I think the provision and download times are particularly important; while I've heard how fast Chassis is to set up and provision, I have virtually never experienced this. Maybe this is because I'm not on fiber internet, I'm currently in Montreal and can expect Chassis to take about 30 minutes to install in total. The chassis repo alone is 30MB!

As much as I too would like to consider this solved and move on from it, I think it's doing harm to the Chassis not to recognize there's a significant problem with the set up times that is a barrier to entry. This is not currently a "light" easy to set up project. Doing a full provision on every install for all my projects leads to massive mental overhead of wanting to set up a new WordPress site (which I think is the opposite of what we want), interruption in my day to take significant work from what I'm trying to focus on and invariably provision errors due to potentially upset package repos, network issues etc.

I'd recommend some solutions to make this better, given it's accepted as a genuine issue that should be addressed:

- A partial / almost fully provisioned base image. This would mean users only ever need download _one_ thing, and then each subsequent chassis setup would re-use the cached image.
- Some way to clone chassis from shared temp folder for performance reasons, so then atleast I'd only ever have to provision everything once.
- Perhaps a way to download a binary with everything bundled in some other way.

rmccue:
My main concern previously was the HDD space that'd be required, however with linked clones in Vagrant 1.8+, this is much better (boxes using a common base box use a differential file system). In addition, with the move to 16.04, we no longer use the `vagrant/precise64` box, so our base box probably isn't installed on most setups any more. These two combined make the argument for our own base box stronger.

What I'd like to do is the following:

1. Introduce tagged versions (#393)
2. Build our own base box with the default configuration.
3. Publish our base box to Vagrant Cloud (presumably `chassis/chassis`).
4. Add a configuration option to use this Chassis base box, rather than the generic one. (For now, this would be off by default, but we would transition this to on by default in a future release.)

With these in place, we can compare the relative time-to-launch for both, including download speeds, and make a judgement as to what we want our default to be.

[...]

PR #468, opened 2017-11-14
Change to the official Chassis base box

BronsonQuick:
**Don't Merge**

This starts work on #390 now that the box has been uploaded to Vagrant Cloud.

I need to test it out more with extensions.


BronsonQuick:
I'm killing this for now as I'm messing around with this on another branch

PR #477, opened 2017-12-11
Enable the /myblog/xmlrpc.php route in multisite mode

jsnajdr:
If I create a multisite box in Chassis (`multisite: Yes`), create a secondary site (`example.com/myblog`) and connect it with Jetpack, then WordPress.com servers will start issuing POST requests to `/myblog/xmlrpc.php`. This route doesn't work in Chassis, however -- it returns a 404.

This patch fixes that by adding a rewrite rule to the nginx config.


rmccue:
Thanks! 🌴

jsnajdr:
Thanks for a quick review @rmccue! 👍

PR #480, opened 2017-12-19
TechCrunch local config setup.

sqsinger1123:
These changes enable SSL to function properly on local development, according to the method of [chassis-openssl](https://github.com/javorszky/chassis-openssl).


sqsinger1123:
My mistake - this branch is only intended for TechCrunch's development. It should not be included in chassis/chassis `master` branch.

PR #500, opened 2018-03-12
Add documenation about disabled_extensions

### LINKED ISSUE #353: Handle deletions of extensions
We should see if we can work out a way to delete extensions. Related: https://github.com/Chassis/phpMyAdmin/issues/14

BronsonQuick:
Fixes #353


BronsonQuick:
>lgtm; do we have docs on handling this from the extension developers' side?

Negative. Good call though! Seeing it kinda varies from extension to extension I might add some basic logic for this in the `example` extension.

I'll update this PR with it in a bit and ask for another review.

PR #504, opened 2018-03-14
Check for updates

### LINKED ISSUE #368: Warn user if submodules are out-of-date (and offer to update for them)
With 2.0, there's major changes to submodules required to provision. When we're provisioning, we should check that submodules are up to date, and offer to run the update for the users, similar to how we initialise them automatically for them.

```
WARNING: Your Chassis submodules appear to be out-of-date, and provisioning may
         fail. Would you like to update them now? [Y/n]
```

BronsonQuick:
This adds functionality to check for updates in Chassis for extensions, submodules and Chassis core. This fixes #368 with bonus features!


BronsonQuick:
I'll need to come back and update this to auto update [global extensions](https://github.com/Chassis/Chassis/pull/575) too

BronsonQuick:
@rmccue I've circled back to this and I've added updates for global extensions too. Can you please give this PR a review? Thanks!

puppet/chassis.rb
```diff
@@ -229,4 +230,103 @@ def self.make_relative(base, paths)
 			pathname.relative_path_from(base).to_s
 		end
 	end
+
+	def self.update_core
+		puts "\e[32mChecking for Chassis core updates...\e[0m"
+		core = ['core']
```
rmccue: Don't really need this var

puppet/chassis.rb
```diff
@@ -229,4 +230,103 @@ def self.make_relative(base, paths)
 			pathname.relative_path_from(base).to_s
 		end
 	end
+
+	def self.update_core
+		puts "\e[32mChecking for Chassis core updates...\e[0m"
```
rmccue: If there's no updates, can we clear this message line to avoid extra output noise? I think it'd be `puts "\r\e[K"`, but not certain.

puppet/chassis.rb
```diff
@@ -229,4 +230,103 @@ def self.make_relative(base, paths)
...
+			next if folder == 'example'
+			if 'core' == folder
+				folder = directory
+				Dir.chdir(directory)
+			elsif
+				Dir.chdir(directory + '/' + folder )
+			end
+			git_checkout_master_stdout, git_checkout_master_stdeerr, git_checkout_master_status = Open3.capture3("git checkout master")
+			git_remote_update_stdout, git_remote_update_stdeerr, git_remote_update_status = Open3.capture3("git remote update")
+			git_status_stdout, git_status_stdeerr, git_status_status = Open3.capture3("git status --porcelain=2 --branch")
```
rmccue: Since we're not using most of these variables, can we just call the function and ignore the return? We should probably check the status, so maybe:
```ruby
_, err, status = Open3...
raise err if status != 0

_, err, status = Open3...
```

(`_` means an unused var in Ruby; it's handled differently by the language IIRC)

puppet/chassis.rb
BronsonQuick: `puts "\033[A\r\e[K"`seemed to do the trick. `\033[A` moves the cursor up one line.

puppet/chassis.rb
BronsonQuick: TIL `_` 🕺

BronsonQuick:
@rmccue Thanks for the review. I've added those changes and I also updated the docs with the new yaml parameters as well. It's ready for another review.

Vagrantfile
```diff
@@ -26,6 +26,10 @@ CONF = Chassis.config
 
 # Install extensions defined in config file
 Chassis.install_extensions(CONF)
+Chassis.update_core(CONF)
```
rmccue: Might be better to have the check here I was thinking? `Chassis.update_core if CONF['auto_update']['core']`

puppet/chassis.rb
```diff
@@ -229,4 +230,124 @@ def self.make_relative(base, paths)
...
+		folders.each do |folder|
+			next if folder == 'example'
+			if 'core' == folder
+				folder = directory
+				Dir.chdir(directory)
+			elsif
+				Dir.chdir(directory + '/' + folder )
+			end
+			_, _, _ = Open3.capture3("git checkout master")
+			_, _, _ = Open3.capture3("git remote update")
```
rmccue: Do we want to check the status return at all? If `checkout` fails e.g. do we want to continue the next commands or report an error and move to the next dir?

BronsonQuick: I just tested if it errors by removing a `.git` folder in one of the extensions and it just continues which is good so I've not bother to return the statuses.

BronsonQuick:
@rmccue Thank you Sir. I've done some more updates. Can you please do another review? Thanks!

rmccue:
> I think this sounds counterintuitive to me; assuming that auto-updates are on by default in a vanilla install, I'd expect `auto_update: false` would default everything, but if I specified `core: false` I'd generally assume that the other keys would be unaffected and remain `true`.

Just to be clear, that's what I meant too. Basically, `$auto_update = merge( $defaults, $prefs )` is what we want (however that works in Ruby)

BronsonQuick:
That's how I've handled it. I did it differently to a merge seeing `auto_update: false` is different to:
```
auto_update:
    core: true
    submodules: true
    extensions: true
    global_extensions: true
```
Let me know if you have any feedback seeing I've made those changes.

BronsonQuick:
>Is this really the right spot to handle global extension updates? If I have one config with global false and one true, I imagine they’d update if I used the VM where it was true; I find it counterintuitive to manage a setting for a global thing within a singular project.

Stellar point Sir! I'll circle back and change that to parse a file from the global extensions directory. 🙇‍♂️

rmccue:
I wonder if (as a broader change) we should have a `~/.chassis/config.yml` as a way to do global defaults. 🤔

BronsonQuick:
>I wonder if (as a broader change) we should have a ~/.chassis/config.yml as a way to do global defaults. 🤔

That makes sense to me then we can do per project overrides. I'll see what I can come up with sometime this week!

BronsonQuick:
@rmccue I've added global updates now so it's ready for another review. Thanks!

BronsonQuick:
@rmccue @kadamwhite Can I please get another review on this too please? 🙇

docs/config.rst
```diff
@@ -429,3 +430,23 @@ Note that if the machine name is changed after it has already been created vagra
...
+
+.. code-block:: yaml
+
+    auto_update:
+    core: true
+        submodules: true
+        extensions: true
+        global_extensions: true
+
+Chassis will automatically check for updates to Chassis core, Chassis core submodules (Apt, MySQL and Stdlib), Chassis extensions and Chassis global extensions. You can stop one of more of the updates from triggering by changing the value(s) to ``false``.
```
tomjn: typo: You can stop one **or** more of

tomjn:
Also, with the bionic beaver changes, for those to take effect will require a `vagrant destroy` and a `vagrant up --provision` so that it uses the new base box, otherwise it'll try to run Bionic provisioners in an Ubuntu 16 VM

BronsonQuick:
@rmccue @roborourke @kadamwhite Ping again for a review please!

puppet/chassis.rb
```diff
@@ -282,4 +292,127 @@ def self.make_relative(base, paths)
...
+			next if folder == 'example'
+			if 'core' == folder
+				folder = directory
+				Dir.chdir(directory)
+			elsif
+				Dir.chdir(directory + '/' + folder )
+			end
+			Open3.capture3("git remote -v update")
+			stdout, _, _ = Open3.capture3("git status --porcelain=2 --branch")
+			if stdout =~ /^# branch.ab \+\d+ -([1-9]\d*)$/
```
roborourke: This check should be simplified down as right now only changes where lines of code have been deleted will register as an update. This can be:
```suggestion
			if stdout !~ /^# branch.ab \+0 -0/
```

BronsonQuick: Good catch Sir! 💖

PR #519, opened 2018-04-09
Bump to PHP 7.2

### LINKED ISSUE #518: Update the default PHP version to 7.2
WordPress 4.9.5 recommends [PHP 7.2](https://github.com/Chassis/Chassis/pull/515/files#diff-2b0386352f41edba269a3ba9befbadb4R61) so we should bump our [version](https://github.com/Chassis/Chassis/blob/master/config.yaml#L103).

BronsonQuick:
Fixes #518.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -100,7 +100,7 @@ synced_folders:
 
 # PHP version
 # Values: 5.6, 7.0, 7.1, 7.2 (or 5.6.30)
-php: 7.0
+php: 7.2
 
 # VirtualBox-specific customisations.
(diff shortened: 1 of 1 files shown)


BronsonQuick:
Now that 4.9.6 has shipped https://wordpress.org/about/requirements/ now recommends PHP 7.2 so we should set that to our default.

rmccue:
I would hold off on this, as PHP 7.2 breaks compat with some older code. Let’s ship with 5.0 instead.

PR #522, opened 2018-04-19
Always restart the service every 30 seconds

### LINKED ISSUE #521: Subdomains Hosts Fail Eventually
I've got subdomains working on a vagrant up, but at some point the hosts mapping stops working. I haven't been able to track down when or why yet, but without fail, the subdomains will stop working after some period of time and use. The main site is still accessible. Also, `/etc/chassis-hosts/conf.d/subdomains` still has all of my subdomains registered in the file.

Running `vagrant reload --provision` will fix them for a short period of time, but `vagrant provision` does not.

1. What operating system do you use?
Mac OSX

2. What version of Vagrant are you running?
vagrant 2.0.3

3. Are you using VirtualBox or VMWare and which version are you using?
Virtualbox 5.2.8

4. Do you have a custom YAML file? If so, what does it contain?

```
# Domain names to use for Vagrant
#
# The first name will be used as the machine's name, subsequent names will be
# aliased in your hosts file and in nginx
#
# (When overriding, include vagrant.local if you need it)
hosts:
  - clientname.local

# Should we use multisite?
multisite: subdomains

# Set database values to save the prefix so Chassis doesn't freak out on
# re-provisions after tests.
database:
    name: wordpress
    user: wordpress
    password: vagrantpassword
    prefix: wp_

# Themes
#
# Install a list of themes, and activate the last one of them
themes:
  - my_theme

# Extensions
#
# Install a list of extensions automatically
extensions:
  - Chassis/Tester
  - Chassis/db-backup
  - Chassis/Xdebug

# PHP version
php: 7.2

# Run database saves on vagrant halts
db-backup:
  halt: yes
  suspend: no
```

5. Do you have any Chassis extensions installed? If so, what are they?

BronsonQuick:
This might fix #521. @mikeselander @joemcgill feel like testing this out for me? You'll need to use this branch and do a `vagrant provision` so the service update is applied.

### DIFF
--- a/puppet/modules/chassis/files/chassis-hosts.service
+++ b/puppet/modules/chassis/files/chassis-hosts.service
@@ -4,7 +4,7 @@ Description=Chassis subdomain hosts updater
 [Service]
 TimeoutStartSec=0
 TimeoutStopSec=600
-Restart=on-failure
+Restart=always
 RestartSec=30s
 StartLimitBurst=5
 ExecStart=/usr/bin/python /vagrant/puppet/chassis-hosts.py


joemcgill:
As mentioned in the original ticket, this didn't seem to fix the issue. And to confirm, I ran a provision and then logged into the box and checked the contents of `/lib/systemd/system/chassis-hosts.service` just to be sure it was actually applied.

BronsonQuick:
@joemcgill Thanks for testing it!

PR #530, opened 2018-06-23
Switch to utf8mb4_unicode_520_ci

### LINKED ISSUE #529: Change DB_COLLATE
We should change `DB_COLLATE` to `utf8mb4_unicode_520_ci` by default. I've notice a few Human Made people needed to switch to `utf8mb4_unicode_ci ` before but as @rmccue noted in an internal repo `utf8mb4_unicode_520_ci` is best.

BronsonQuick:
Fixes #529.

I've tested this with a fresh Vagrant up and verified the collation using Sequel Pro.

### DIFF
--- a/wp-config.php
+++ b/wp-config.php
@@ -98,7 +98,7 @@ if ( WP_LOCAL_DEV && ! defined('WP_SITEURL') && ! defined( 'WP_INSTALLING' ) ) {
 // You almost certainly do not want to change these
 // ================================================
 define( 'DB_CHARSET', 'utf8' );
-define( 'DB_COLLATE', '' );
+define( 'DB_COLLATE', 'utf8mb4_unicode_520_ci' );
 
 // ==============================================================
 // Salts, for security


rmccue:
I think it's actually better to leave this empty, as it will then use WP's defaults: https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/wp-db.php#L767-L774

rmccue:
@ntwb `utf8mb4_unicode_520_ci` will be used for `utf8mb4`; the [check that activates it](https://github.com/WordPress/WordPress/blob/6fd8080e7ee7599b36d4528f72a8ced612130b8c/wp-includes/wp-db.php#L777) checks that mysql has `utf8mb4_520` is available, not that the charset is that.

The optimal combination is `utf8mb4` for the charset and `utf8mb4_unicode_520_ci`, which are both triggered by just setting `DB_CHARSET` to `utf8` and not setting `DB_COLLATE` (same as currently set up).

BronsonQuick:
Thanks gents! Looks like I must've missed some details on that client project seeing we needed to add this into our setup for that. I'll close this one out then!

PR #534, opened 2018-06-28
Remove old PHP references

### LINKED ISSUE #511: Remove references and setup for Php versions less than 5.6
Since the PPA for php-oldstable was removed we can no longer support versions of Php less than 5.6. We still have code to handle setup of this. e.g. https://github.com/Chassis/Chassis/blob/9bfecaf9b4e9f97910dac92db97248e735443f08/puppet/modules/chassis/manifests/php.pp#L35

https://github.com/Chassis/Chassis/blob/master/puppet/modules/chassis/files/php-5.3.init

We should do an audit and remove all those old references.

BronsonQuick:
Fixes #511


puppet/modules/chassis/manifests/php.pp
```diff
@@ -131,8 +89,6 @@
 					ensure => absent,
 				}
 			}
-			# "5.3",
-			# "5.4",
 			default: {
```
rmccue: We can actually remove this `case` now, as all versions now fall under the other case.

PR #536, opened 2018-07-02
Make the folder names lowercase

### LINKED ISSUE #535: Make the folder names for extensions lowercase
I've heard of a few people hitting issues with some projects where someone has added a Chassis extension like so:
```
extensions:
 - Tester
```
on one branch and another developer has added:
```
extensions:
 - tester
```
on another branch and then the project ends up with an extension in a `Tester` folder as well as a `tester` folder then Chassis won't `vagrant up` or `vagrant provision`.

We can easily fix this by forcing the git clone into a folder with a lowercase name.

BronsonQuick:
Fixes #535

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -158,8 +158,8 @@ module Chassis
 			repo = extension
 		end
 
-		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '')
+		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '').downcase
 
-		system("git clone %s %s --recursive" % [repo, folder.downcase] ) unless File.exist?( folder )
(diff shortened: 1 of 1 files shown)


puppet/chassis.rb
```diff
@@ -160,6 +160,6 @@ def self.install_extension(extension)
 
 		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '')
 
-		system("git clone %s %s --recursive" % [repo, folder] ) unless File.exist?( folder )
+		system("git clone %s %s --recursive" % [repo, folder.downcase] ) unless File.exist?( folder )
```
rmccue: This should be on the `folder =` line, otherwise `File.exist?` will fail

BronsonQuick: Good catch. I've altered that now. Thanks!

PR #548, opened 2018-07-09
Increase client_max_body_size size to match php

### LINKED ISSUE #547: Increase client_max_body_size in nginx to 100M
Currently we set both [post_max_size](https://github.com/Chassis/Chassis/blob/master/puppet/modules/chassis/templates/php.ini.erb#L675) and [upload_max_filesize](https://github.com/Chassis/Chassis/blob/master/puppet/modules/chassis/templates/php.ini.erb#L803) to 100M but we're setting [client_max_body_size](https://github.com/Chassis/Chassis/blob/master/puppet/modules/chassis/templates/site.nginx.conf.erb#L5) in Nginx to 50M. This means if a user uploads a file greater than 50M and less than 100M they'll receive a HTTP Error which is a 413.

BronsonQuick:
Fixes #547.

### DIFF
--- a/puppet/modules/chassis/templates/site.nginx.conf.erb
+++ b/puppet/modules/chassis/templates/site.nginx.conf.erb
@@ -2,7 +2,7 @@ server {
 	listen 80;
 	server_name <%= @server_name %>;
 	root <%= @location %>;
-	client_max_body_size 50M;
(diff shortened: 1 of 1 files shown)


BronsonQuick:
Thanks @ntwb! 💖

PR #550, opened 2018-07-13
Do not install plugins or themes until WordPress is available.

### LINKED ISSUE #549: Plugins not installed on initial provision
1. OS / Vagrant / VBox versions: **Ubuntu 18.04** / **Vagrant 2.2.1** / **VirtualBox 5.2.10**

2. Do you have a custom YAML file? yes:

```
plugins:
  - jetpack
```

3. No chassis extensions installed.

-----------

**Expected**

When cloning Chassis, creating a `content/config.local.yaml` file specifying one or more `plugins:`, and running `vagrant up`, I expect that plugins will be downloaded and activated during that initial provision.

**Actual**

After the initial `vagrant up`, the `content/plugins` directory is empty, and no plugins have been downloaded or activated. I must run `vagrant provision` once more to install plugins.

The last lines of the provisioner output from the `vagrant up` read:
```
==> default: Notice: /Stage[main]/Main/Chassis::Wp[test.local]/Chassis::Site[test.local]/Wp::Site[/vagrant/wp]/Exec[wp install /vagrant/wp]/returns: executed successfully
==> default: Notice: Finished catalog run in 144.29 seconds
==> default: Running provisioner: shell...
    default: Running: /tmp/vagrant-shell20180713-30872-ba4082.sh
    default: ======================================
    default: Your Chassis box has been provisioned!
    default: ======================================
    ...(domain, login credentials, etc, then process ends)
```
When running `vagrant provision` after that, I now see
```
==> default: Notice: /Stage[main]/Main/Chassis::Wp[test.local]/Wp::Plugin[jetpack]/Exec[wp install plugin jetpack --activate]/returns: executed successfully
```

I'm guessing this is a result of the fact that wp-cli is used to install the plugins, so it cannot `install --activate` unless the site is installed. I have been poking around in the `.pp` files but I'm not sure where to look to validate or address this possible order-of-operations issue.

kadamwhite:
`wp::plugin` & `wp::theme` run `onlyif  => "/usr/bin/wp core is-installed"`, so because the `chassis::wp` def doesn't do any explicit task sequencing these tasks run before `chassis::site` & silently fail on `vagrant up`. Plugins & Themes will only be installed on subsequent provisions.

To fulfill the expectation that the state described in your yaml config describes the state the box will be on after the initial provision, this PR breaks the `wp::plugin` and `wp::theme` calls into a separate dev which gets explicitly sequenced after `chassis::wp`. This ensures that the WordPress environment is available when we attempt installation, allowing plugins and themes to be pulled down in the initial `vagrant up`.

Fixes #549

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -55,14 +55,20 @@ chassis::wp { $config['hosts'][0]:
 	admin_user        => $config[admin][user],
 	admin_email       => $config[admin][email],
 	admin_password    => $config[admin][password],
-	plugins           => $config[plugins],
-	themes            => $config[themes],
 
 	extensions        => $extensions,
 
-	require  => [
+	require => [
 		Class['chassis::php'],
 		Package['git-core'],
 		Class['mysql::server'],
 	]
+}
+
+# Tasks wp::plugin and wp::theme run onlyif "/usr/bin/wp core is-installed":
+# sequence chassis::content after chassis::wp.
+-> chassis::content { $config['hosts'][0]:
+	location => $config[mapped_paths][base],
+	plugins  => $config[plugins],
+	themes   => $config[themes],
 }
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -65,10 +65,11 @@ chassis::wp { $config['hosts'][0]:
 	]
 }
 
-# Tasks wp::plugin and wp::theme run onlyif "/usr/bin/wp core is-installed":
-# sequence chassis::content after chassis::wp.
--> chassis::content { $config['hosts'][0]:
+chassis::content { $config['hosts'][0]:
 	location => $config[mapped_paths][base],
 	plugins  => $config[plugins],
 	themes   => $config[themes],
+
+	# These tasks will not run unless WP is installed.
+	require  => Chassis::Wp[ $config['hosts'][0] ],
 }
--- a/puppet/modules/chassis/manifests/content.pp
+++ b/puppet/modules/chassis/manifests/content.pp
@@ -1,0 +1,16 @@
+# Install our WordPress site and add our configuration.
+define chassis::content (
+	$location,
+	$plugins = [],
+	$themes = [],
+) {
+  wp::plugin { $plugins:
+    ensure   => 'enabled',
+    location => $location,
+  }
+
+  wp::theme { $themes:
+    ensure   => 'enabled',
+    location => $location,
+  }
+}
--- a/puppet/modules/chassis/manifests/content.pp
+++ b/puppet/modules/chassis/manifests/content.pp
@@ -1,4 +1,4 @@
-# Install our WordPress site and add our configuration.
+# Install WordPress themes and plugins.
 define chassis::content (
 	$location,
 	$plugins = [],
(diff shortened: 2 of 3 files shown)


puppet/manifests/development.pp
```diff
@@ -55,14 +55,20 @@
...
+	require => [
 		Class['chassis::php'],
 		Package['git-core'],
 		Class['mysql::server'],
 	]
 }
+
+# Tasks wp::plugin and wp::theme run onlyif "/usr/bin/wp core is-installed":
+# sequence chassis::content after chassis::wp.
+-> chassis::content { $config['hosts'][0]:
```
kadamwhite: It's possible that there's a way to use `require` in place of `->` here, which might be conceptually preferable, but this works and I burned too much time trying to understand puppet defs _vs_ classes to continue pursuing that approach. :)

kadamwhite:
If there is a better way to structure this such that other extensions could then know when WordPress is installed, that is of extreme interest to me: I am struggling in another extension that I am trying to write to know how to make sure that I do not run my (WP-CLI) commands until WordPress is available

kadamwhite:
@bronsonquick thanks for the review! Do you know how I might be able to author an extension that would run a wp-cli command only post-wp install?

BronsonQuick:
Ahh good question. That's not something I've had to do before for an extension. Maybe a `require => Class['wp::cli']` in your extension might work? You might also need to do an `include wp::cli` in your extension as well?

puppet/manifests/development.pp
rmccue: We generally only use `require`. For this, it will be:
```
require => Chassis::Wp[ $config['hosts'][0] ],
```

puppet/modules/chassis/manifests/content.pp
```diff
@@ -0,0 +1,16 @@
...
+  wp::plugin { $plugins:
+    ensure   => 'enabled',
+    location => $location,
+  }
+
+  wp::theme { $themes:
+    ensure   => 'enabled',
+    location => $location,
+  }
+}
```
rmccue: This file should use tabs rather than spaces.

puppet/manifests/development.pp
rmccue: btw, the rule for referencing is:
* Take the declaration line which looks like `foo::bar { "quux":`
* Split the first bit (the *resource type*) by namespace separator (`::`)
* Uppercase the first letter of each piece
* Combine together with namespace separator
* Add the second bit (the *resource name*) in square brackets

So, `class { 'foo':` -> `Class['foo']`, `chassis::wp{ $x:` -> `Chassis::Wp[ $x ]`

kadamwhite:
@rmccue @BronsonQuick Thank you both for the review! And thanks especially Ryan for the syntax explanation. It makes me feel a little better about the time I lost this weekend punching at a brick wall -- turns out I got _very_ close, but didn't know the capitalization thing!

I've re-done the patch to use `require` instead of `->`, and fixed the indentation problem. Back to y'all for review.

BronsonQuick:
I've just done a fresh clone and a `vagrant up` just to test we're all good to merge this and we're all good! Thanks folks. Much appreciated!
chassis__fish__volumessiteschassis__-fish_2018-07-17_14-26-19 1

PR #553, opened 2018-07-17
Add support for PHP extensions and ini overrides

### LINKED ISSUE #551: Make PHP ini file configuration part of core
The extension over at https://github.com/Chassis/phpini is really useful, but I think given that PHP is provisioned and part of the core chassis set up it might make sense to bake that functionality in.

It'd be awesome to be able to do this via config too to reduce the number of dev specific committed files and allow local overrides eg:

```yaml
php-ini:
  memory_limit: 256M
```

roborourke:
Allows setting PHP extensions and overrides using the following syntax:

```
php_config:
  ini:
    memory_limit: 256M
  extensions:
    - intl
```

Fixes #551

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -63,6 +63,29 @@ You can use either a two-part version (``5.3``) or a three-part version
 .. note::
    There are a few known issues on Windows with PHP 5.4.
 
+PHP Configuration
+-----------------
+
+**Key**: ``php_config``
+
+While Chassis provisions PHP with the necessary packages and standard configuration
+needed to run WordPress these might not always match the set up you have in
+production, or you may want to modify certain configuration settings to test against.
+
+To add extra extensions such as `intl` or override the default `php.ini`
+settings add a section to your config file like so::
+
+   php_config:
+      extensions:
+         - intl
+         - pdo
+         - imagick
+      ini:
+         memory_limit: 256M
+         post_max_size: 10M
+
+.. note::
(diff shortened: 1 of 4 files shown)


roborourke:
Let me know if you think this might be better split out, eg. as `php_ini` and `php_extensions` keys

roborourke:
@rmccue cheers for taking a look.

I would've ideally like to use the `php` key and provide the settings from that as a hash but it would be a bit of a nightmare for back compat.

Noted re. having a separate ini file too, that makes sense.

I wasn't sure how best to handle the extensions using the existing parameter because they're hardcoded in `development.pp` right now. Can `union` them in that pp file before passing them instead.

rmccue:
> I would've ideally like to use the `php` key and provide the settings from that as a hash but it would be a bit of a nightmare for back compat.

That can potentially be handled in the config parser in Ruby instead if you _really_ want to, but I would say it's probably not a great way to do it.

> I wasn't sure how best to handle the extensions using the existing parameter because they're hardcoded in `development.pp` right now. Can `union` them in that pp file before passing them instead.

Yeah, I would essentially take the static list we have and merge it with one from the config.

BronsonQuick:
I'm closing this as there was a discussion about this in #https://github.com/Chassis/Chassis/issues/551

roborourke:
Cheers @BronsonQuick. I'll look at sorting out the modification for installing PHP extensions in another PR.

PR #563, opened 2018-07-24
Add instructions for Git URLs

### LINKED ISSUE #239: Support for pulling themes and plugins from a repo
If possible, running theme and plugin installs set in config.yaml that are sourced from a git repository

BronsonQuick:
Fixes #239.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -310,6 +310,8 @@ Plugins
 If you're using plugins from the WordPress.org repository you can add them in a list using the plugins slug.
 These will be downloaded, installed and activated for you.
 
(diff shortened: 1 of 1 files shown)


docs/config.rst
```diff
@@ -326,13 +329,16 @@ Themes
 If you're using themes from the WordPress.org repository you can add them in a list using the themes slug.
 These will be downloaded for you. The last theme in the list will be the theme that is activated for your site.
 
+Alternatively, if you want to install a plugin from a Git repository you can use a URL to a zip file of your theme. e.g. ``https://github.com/humanmade/S3-Uploads/archive/master.zip``
```
ntwb: Did you want to use `theme` here instead of `plugin` here? i.e `Alternatively, if you want to install a theme...`

BronsonQuick: I sure do! Good catch. Thanks!

BronsonQuick:
Updated!

PR #564, opened 2018-07-30
Custom default domain

### LINKED ISSUE #496: Change nginx's default domain page to custom error page
If you hit Chassis with a host it doesn't know about in single-site mode, it gives you the default "Welcome to nginx!" page. We should customise this with detail about why people might be hitting it.

BronsonQuick:
Fixes #496.

If you want to short circuit this for testing you can comment out the[ following lines](https://github.com/Chassis/Chassis/blob/master/puppet/manifests/development.pp#L44-#L75).

@rmccue Can you please have a good read over this as my English tends to be shocking and it'll no doubt require some corrections. Also, please let me know if you can think of any other troubleshooting ideas and I'll add them in. Thanks!

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -73,3 +73,9 @@ chassis::content { $config['hosts'][0]:
 	# These tasks will not run unless WP is installed.
 	require  => Chassis::Wp[ $config['hosts'][0] ],
 }
+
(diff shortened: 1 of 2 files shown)


BronsonQuick:
@rmccue Another annoying ping for this one Sir! 🙇‍♂️

BronsonQuick:
@rmccue Woot! Thanks man. 💖

PR #570, opened 2018-08-28
Make the WordPress site title configurable through yaml

kadamwhite:
While working on client projects it can be a nice touch to ensure a new VM has the appropriate site title. There was a line of code in chassis which seemed to try to set the installed name to 'Vagrant Site,' but it was using the wrong wp::site parameter and it did not allow the value to be configured. This patch adds a new `site` key to the YAML file, permitting users to specity a `name` value to use as the site title for their new VM.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,6 +17,11 @@
 hosts:
     - vagrant.local
 
+# Site Configuration
+# (When overriding, include all values)
+site:
+    name: Chassis Site
+
 # IP address to use on the private network
 #
 # DHCP is used by default to assign IP addresses dynamically, override only if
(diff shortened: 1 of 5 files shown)


kadamwhite:
@rmccue AFAIK configuring multi-site domains is complex enough that we tend to need to write a custom script anyway; I'd love better support for it in Chassis but I'd think it would be a good follow-up ticket rather than an add-on to this patch. If there's a more future-proof key to use in the config though, I'm happy to defer to your judgement.

kadamwhite:
@BronsonQuick @rmccue What more is needed here to move this forward?

BronsonQuick:
@kadamwhite Can you please add:
```
# Site Configuration
# (When overriding, include all values)
site:
    name: Chassis Site
```
to `config.yaml` in the root directory and this will be good to merge.

kadamwhite:
@BronsonQuick done! Back to you.

PR #575, opened 2018-09-09
Add global extensions

### LINKED ISSUE #574: Add support for global extensions
There are basically two types of extensions that I've seen:

1. Server-related packages, which are required for the site to run. These are things like specific services (Cavalcade, etc) or PHP extensions (v8js).
2. Development tools, which are mostly up to personal preference. These are things like XDebug, MailHog, SequelPro, db-backup.

While our configuration allows a split between local preferences and committed preferences, no such split exists for extensions. In addition, the development tools are typically things you want on all of your Chassis boxes, rather than replicating every time.

To that end, I think we should add global extensions. These would simply be extensions you place in a system-wide place (i.e. `~/.chassis/extensions`) that Chassis automatically adds to every box for you. This would allow you to ensure that your tools are always available regardless of the project's configuration. (However, if there are duplicate extension names, the one in the project should always take precedence.)

rmccue:
Fixes #574.

To test: create a `~/.chassis/extensions` directory, and clone an extension (e.g. MailHog) into it. Then provision your box, and note that MailHog is provisioned on it.

Remaining items:
* [x] Write documentation

### DIFF
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -30,6 +30,21 @@ To install the memcached extension you would do the following:
 3. `vagrant provision`.
 
 
+Globally installing extensions
+------------------------------
+
+In addition to per-project, extensions can be globally installed. This is useful
+for development tools that you want to apply to all your Chassis boxes, which
+aren't specific to any project.
+
+Chassis loads global extensions from ``~/.chassis/extensions`` in the same way
+that it loads project-specific extensions.
+
+If a project includes an extension of the same name in its ``extensions/``
+directory, the project's extension will be loaded instead of the
+globally-installed extension.
+
+
 Creating your own
 -----------------
 
--- a/docs/extend.rst
+++ b/docs/extend.rst
@@ -21,7 +21,7 @@ as extensions, so extensions can be downloaded into this directory.
 
 
 Example: Installing The Memcached Extension
--------------------------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To install the memcached extension you would do the following:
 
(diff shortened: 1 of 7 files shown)


docs/extend.rst
```diff
@@ -30,6 +30,21 @@ To install the memcached extension you would do the following:
...
+Globally installing extensions
+------------------------------
+
+In addition to per-project, extensions can be globally installed. This is useful
+for development tools that you want to apply to all your Chassis boxes, which
+aren't specific to any project.
+
+Chassis loads global extensions from ``~/.chassis/extensions`` in the same way
+that it loads project-specific extensions.
+
```
BronsonQuick: I'm thinking we should probably add an example for reference. e.g. `git clone https://github.com/Chassis/Mailhog ~/.chassis/extensions/mailhog`

puppet/chassis.rb
```diff
@@ -20,17 +20,28 @@ def self.get_extension_config(extension)
...
+			config = get_extension_config(extension, ext_dir)
 			config['version'] == version
 		}
 	end
 
+	def self.get_extensions(version = nil)
+		self.get_extensions_for_dir(@@extension_dir, version)
+	end
+
+	def self.get_global_extensions(version = nil)
```
BronsonQuick: I don't think you're calling this function anywhere at the moment?

rmccue: Yeah, I added this for extensions that want to use it, but we don't use it internally.

BronsonQuick:
I'm going to throughly test this out after yesterdays changing. I'll be working my way through the following tests before we merge this:
- [ ] No extensions
- [ ] One project extension
- [ ] One global extension
- [ ] One global and one project extension
- [ ] Two global extensions and no project extensions
- [ ] Two global extensions and two project extensions

Issue #581, opened 2018-09-15
Configuration files in non-standard `content` directory ignored


mikeselander:
When using a non-standard content directory, Chassis configuration files in said directories are not parsed, and the configuration files are only evaluated against the default content folder.

Consider the following project setup (common hm-base type setup):
```
 - /chassis (Everything Chassis)
   - config.local.yaml (Basic overrides (see full file below)
 - /content (Actual used content directory)
   - config.local.yaml (Installation-specific overrides)
 - /wordpress (WordPress used on platform)
```

and the following paths:
```yml
paths:
  base: ..
  content: content
  wp: wordpress
```

If one were to run an installation with the above configuration files, the `/content/config.local.yaml` would get ignored. It's only after adding a `/chassis/content` directory and adding the config file to _that_ directory that the customizations would take effect.

I would expect that Chassis would take a look at `chassis/config.local.yaml` (since this should always be in the same place, it seems, and if there is a custom content path set, load the _custom_ content directory configuration files as opposed to the default Chassis one.

1. What operating system do you use?
Mas OS X 10.13.6

2. What version of Vagrant are you running?
2.1.5

3. Are you using VirtualBox or VMWare and which version are you using?
Virtualbox 5.2.18

4. Do you have a custom YAML file? If so, what does it contain?
```
# Establish our custom paths for local.
paths:
  base: ..
  content: content
  wp: wordpress

# Default domain to known.
hosts:
  - client.local

# Define database prefix, since we have tests as well.
database:
  name: wordpress
  user: wordpress
  password: vagrantpassword
  prefix: wp_

# Install a list of Chassis extensions automatically.
extensions:
  - Chassis/Cavalcade
  - Chassis/Debugging
  - Chassis/Memcache
  - Chassis/SequelPro
  - Chassis/Tester
  - Chassis/Xdebug

# PHP version.
php: 7.0

# Setup a predictable administrator to begin with.
admin:
  user: clientdev
  email: admin@example.com
  password: password

# Default our theme activated and turned on.
themes:
  - client
```

5. Do you have any Chassis extensions installed? If so, what are they?
*See above*

rmccue:
There’s a few reasons we don’t do this right now:

1. It would require essentially a two pass configuration parsing, as we would need to read and combine the default files first, then if a custom content dir is set, reparse using the first config parse.
2. The `paths` directive itself would need to be special-cased, as it can only really be set in the main directory. This is likely tough to explain in the documentation.
[...]

PR #583, opened 2018-09-22
Allow adding Chassis as a Composer dependancy

### LINKED ISSUE #481: Allow adding Chassis as an npm/Composer dependancy
This is more of a thought experiment. It would be great to allow adding Chassis as an npm or Composer dev dependancy:

    $ npm install chassis/chassis --save-dev

or 
   
    $ composer require --dev chassis/chassis

This requires the following changes:

1. A very minimal `package.json` and `composer.json` in the Chassis root directory.

2. Update `self.load_config` in `puppet/chassis.rb` to look for `config.yaml` overrides in the project root directory in addition to the [Chassis root directory `@@dir`](https://github.com/Chassis/Chassis/blob/c187e6e5c9be9f66322b683025a66af541181f58/puppet/chassis.rb#L5).

3. Update `self.normalize_config` to accept `base` outside of the Chassis directory.

kasparsd:
Fixes #481.

## Objectives

Allow adding Chassis to any project as a Composer or npm dependancy:

```
composer require --dev chassis/chassis
```

with only two extra files:

`Vagrantfile` that loads the Chassis `Vagrantfile` and specifies the project root as the directory that contains the configuration file:

```
ENV['CHASSIS_CWD'] = File.dirname(__FILE__)
load "vendor/chassis/chassis/Vagrantfile"
```

and `config.local.yaml` that adds the project directory as a plugin or theme to Chassis:

```yaml
hosts:
  - composer.local

synced_folders:
  .: /vagrant/content/plugins/this-plugin
```

## Suggested Changes

- Always map Chassis root directory to the `/vagrant` directory inside the Vagrant machine.

- Map the Vagrant working directory (directory of `Vagrantfile`) **or** `ENV['CHASSIS_CWD']` ([see discussion](https://github.com/Chassis/Chassis/issues/481#issuecomment-423943722)) to `/chassis` inside Vagrant.

- Always look for config override files `config.local.yaml` in `/chassis`.

By default, the Vagrant working directory is the same as the Chassis root directory so `/chassis` will map to to the same place as `/vagrant`. However, in case of adding Chassis via Composer or npm, the `/chassis` directory will map to the root of that project while `/vagrant` will still map to the root of the Chassis core.

Settings for `paths` in `config.local.yaml` are still resolved relative to the Chassis root directory, in order to preserve the default lookups for the `wp` and `contents` directories. 

## Tasks

- [x] All Chassis system files should be loaded relative to the Chassis root directory.
- [x] Allow `config.yaml` overrides in a directory outside the Chassis root.

### Example: Plugin Development

See https://github.com/wpsh/chassis-composer-demo/tree/dev-481

- Example of a plugin with an included development environment via Composer.
- Installs the Query Monitor plugin for development purposes.
- Uses `synced_folders` to define the project root directory as a plugin mapped to `/chassis/content/plugins/this-plugin`.

```bash
git clone https://github.com/wpsh/chassis-composer-demo.git
cd chassis-composer-demo
git checkout dev-481
composer install
vagrant up
```

### Example: WP Core Development

See https://github.com/wpsh/chassis-composer-demo/tree/dev-481-wp-develop

- Adds [`wordpress-develop`](https://github.com/WordPress/wordpress-develop) as a Composer dependancy.

## Known Issues

- None.

## Notes

See: https://kaspars.net/blog/wordpress/wordpress-environment-composer


BronsonQuick:
Thanks so much for starting work on this @kasparsd! Let me know when you're happy for one of us to test it out for you.

Vagrantfile
```diff
@@ -9,11 +9,22 @@ if Gem::Version.new(Vagrant::VERSION) < Gem::Version.new("1.5.0")
...
+chassis_config_dir = chassis_dir
+
+# Load Chassis config overrides from a custom directory, if configured.
+unless ENV['CHASSIS_CWD'].nil?
+	chassis_config_dir = ENV['CHASSIS_CWD']
+end
+
 # Check that submodules have been loaded
-if not File.exist?(File.join(File.dirname(__FILE__), "puppet", "modules", "apt", ".git"))
+if not File.exist?(File.join(chassis_dir, "puppet", "modules", "apt", ".git"))
```
kasparsd: `File.dirname(__FILE__)` was repeated in several places so we introduce `chassis_dir`.

kasparsd:
@BronsonQuick This is ready for code review and testing. See the Examples section in the description for two sample repositories.

rmccue:
@kasparsd I think you've potentially mixed two different directories here? Extensions/Puppet/etc should _always_ be at `/vagrant/extensions`, `/vagrant/puppet`, etc, I don't think we want to change that. (We could potentially change to always mount `root` at `/chassis` though.)

Perhaps I'm missing what this gains over the existing `VAGRANT_CWD`?

kasparsd:
@rmccue Thanks for reviewing the pull request!

> Extensions/Puppet/etc should always be at /vagrant/extensions, /vagrant/puppet, etc.

Just to confirm -- could we mount Chassis root at `/vagrant` and the project root at `/chassis`? I'll look into this.

> Perhaps I'm missing what this gains over the existing VAGRANT_CWD?

I created [this sample project](https://github.com/wpsh/chassis-composer-demo/tree/master) to illustrate the issue.

With `VAGRANT_CWD=vendor/chassis/chassis vagrant up` it looks for config overrides `config.local.yaml` inside the Chassis root directory `vendor/chassis/chassis` instead of the project directory. This is due to [this lookup logic](https://github.com/Chassis/Chassis/blob/c187e6e5c9be9f66322b683025a66af541181f58/puppet/chassis.rb#L59) which assumes that all config overrides are at the Chassis root `@@dir`. 

The suggested fix is to introduce a `@@config_dir` variable which can be set to anything via the `CHASSIS_CWD` environmental variable. We need some way of telling Chassis that it should read the config override from another directory.

.gitignore
```diff
@@ -6,12 +6,11 @@
 /local-config-extensions.php
 /sql-dump-*.sql
 /db-sync
-/content
 .DS_Store
 .idea
 config.local.yaml
 /.wp-cli
 /wp-cli.local.yml
-/extensions
```
kasparsd: This directory is actually being tracked as [it contains the `example` extension](https://github.com/Chassis/Chassis/tree/c187e6e5c9be9f66322b683025a66af541181f58/extensions/example).

.gitignore
```diff
@@ -6,12 +6,11 @@
 /local-config-extensions.php
 /sql-dump-*.sql
 /db-sync
-/content
```
kasparsd: We actually need this directory to exist so we add a dedicated `.gitignore` inside the `content` directory which excludes everything but itself.

puppet/chassis.rb
```diff
@@ -3,6 +3,7 @@
 
 module Chassis
 	@@dir = File.dirname(File.dirname(__FILE__))
+	@@config_dir = @@dir
```
kasparsd: This is the new config path variable used to reference the directory that contains the config overrides. Previously it was assumed that all config overrides are inside the Chassis root directory.

.gitignore
rmccue: This is intentional; `gitignore` simply ensures untracked files are not shown, but still allows files to be committed within it. Chassis requires the `example` extension for technical reasons (plus it's useful), but we don't want to show any other files in that directory.

.gitignore
rmccue: This will conflict with anyone with a checked-out repository inside `content`, and is a blocker.

puppet/chassis.rb
```diff
@@ -93,23 +94,11 @@ def self.normalize_config(config)
...
-			rel_path = base_path.relative_path_from( Pathname.new @@dir )
-			if rel_path.to_s != "."
-				config["mapped_paths"]["base"] = "/vagrant/" + rel_path.to_s
-			else
-				config["mapped_paths"]["base"] = "/vagrant"
-			end
-		end
+		# Set up the paths as needed
+		config["mapped_paths"] = {}
+		config["mapped_paths"]["base"] = "/vagrant"
```
rmccue: This still appears to be using the incorrect split of base/Chassis directories?

rmccue:
I wonder whether we could try and address both this issue and #581 at the same time.

What if instead of a custom environment variable for this, we instead loaded `config{,.local}.yaml` from the specified content directory? This would require a two-pass parse, but would be a bit nicer IMO.

.gitignore
kasparsd: @rmccue Right, so we want to keep tracking `/extensions` directory and the `example` directory inside it while ignoring everything else. This is why I suggest we introduce a `.gitignore` in the `extensions` directory` that only tracks the `example` directory and excludes everything else.

Currently it is a bit confusing to see the `extensions` directory ignored in the `.gitignore` but actually tracked in the repository. I think it only works because it was once tracked and later ignored in 9b49b5637a6b614a4d16ff1320aed80f38aa6105. Per [docs](https://git-scm.com/docs/gitignore):

> Files already tracked by Git are not affected.

and

> To stop tracking a file that is currently tracked, use `git rm --cached`.

So if the intention is to track the sample extension `extensions/sample`, I suggest we clarify that by updating the `.gitignore` files.

I'm totally OK with not changing this, if you prefer. Just wanted to explain the reasoning and though behind it.

Vagrantfile
```diff
@@ -9,31 +9,49 @@ if Gem::Version.new(Vagrant::VERSION) < Gem::Version.new("1.5.0")
...
 		puts "WARNING: Submodules may be missing, and could not automatically\ndownload them for you."
 	end
 
 	# Extra new line, please!
 	puts
 end
 
+# Ensure we have a content directory to sync.
+chassis_content_dir = File.join(chassis_dir, "content")
+Dir.mkdir(chassis_content_dir) unless File.exists?(chassis_content_dir)
```
kasparsd: @rmccue We need the `content` directory to exist before Vagrant can sync it to the virtual machine or it fails with the following error on the initial `vagrant up`:

```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

vm:
* The host path of the shared folder is missing: /absolute/path/to/Chassis/content
```

kasparsd:
Closing since I've started working on https://github.com/wpsh/wpsh-local

PR #600, opened 2019-01-09
Update apt automatically

### LINKED ISSUE #582: Update Apt automatically
As a developer, my apt packages should update frequently otherwise I run the risk of bumping a PHP version in my configuration file and I can get a failed provision.

I have had code for this here: https://github.com/Chassis/Chassis/pull/476/files#diff-556e699d648d31f209eec94844939770R15 it's important that we ship this for developers.

BronsonQuick:
Fixes #582.

Although the order of this looks weird, if this is added after `include apt` the provisioning fails!

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -20,12 +20,10 @@ if $loadable_extensions {
 }
 
 class { 'apt':
-  update => {
-	  frequency => 'weekly',
-  },
+	update => {
+		frequency => 'weekly',
+	},
 }
-
-include apt
 
 class { 'chassis::php':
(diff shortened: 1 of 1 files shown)


rmccue:
> Although the order of this looks weird, if this is added after `include apt` the provisioning fails!

I think we should remove the `include` line altogether; IIRC, `include x` is just a shortcut for `class { 'x': }` anyway. Per [the docs](https://puppet.com/docs/puppet/5.3/lang_classes.html#include-like-vs-resource-like) you're not meant to have both.

BronsonQuick:
@rmccue Ahh thanks for that. I've removed that include now!

PR #614, opened 2019-04-13
Add some proxy settings for Johnny B

### LINKED ISSUE #585: Increase nginx's proxy timeout
When stepping through breakpoints in PHP, the proxy timeout often gets hit, causing nginx to respond with a 504.

It would be great if the default proxy timeout was significantly increased in order to avoid this problem.

BronsonQuick:
I don't know how to replicate it but I think this might fix #585.


BronsonQuick:
@johnbillion Thank you Sir! 😍

PR #586, opened 2018-10-26
Handle global extensions that might have .vagrantplugins as well

BronsonQuick:
Fixes #576

PR #616, opened 2019-04-29
Fix empty check

rmccue:
Turns out this should have been `.empty?` instead of `.empty` 🤦‍♂ 

See #586.


rmccue:
YOLO-merging, as this is causing new installs to fail.

PR #621, opened 2019-04-30
Add explicit requirement on local-config files

### LINKED ISSUE #620: Add explicit dependency between local-config-*.php and chassis::wp
We don't currently have any hard dependency between `chassis::network`/`chassis::site` inside `chassis::wp` and the `local-config-{db,extensions}.php` files. It's basically just luck that these work!

rmccue:
Fixes #620. I swear I've hit every edge case in Chassis in the past day.

### DIFF
--- a/puppet/modules/chassis/manifests/wp.pp
+++ b/puppet/modules/chassis/manifests/wp.pp
@@ -33,6 +33,10 @@ define chassis::wp (
 			admin_user        => $admin_user,
 			admin_email       => $admin_email,
 			admin_password    => $admin_password,
+			require           => [
+				File['/vagrant/local-config-db.php'],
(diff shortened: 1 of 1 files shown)


rmccue:
The install was failing with a 255 code, as our custom config had `require local-config-extensions.php` (Chassis only has `include`). I noticed that the file was only getting provisioned after the install command when this happened. Total luck that `local-config-db.php` was getting provisioned before it, otherwise installs would have always failed.

BronsonQuick:
Ahh right. Thanks for the update Sir! The `255` is so common when things go pear shaped with Chassis with some extensions on `vagrant up`. The most common one that springs to mind is memcached!

rmccue:
Yeah, 255 is just the generic "PHP fatal error" return code

PR #629, opened 2019-05-21
Make sure MULTISITE is always true

### LINKED ISSUE #615: Can't activate user's account in a multisite
I wanted to test a multisite in Chassis and I was able to enable it but I have a problem with activating users' accounts.

As a super admin I allowed site administrators to add new users. Then as a site administrator I added a new user and I got an activation email. I opened the email in MailHog and tried to click the activation link which looks like this `http://wordpress.test/site3/wp-activate.php?key=b5a5f9162d8097f9` but I get the error `Registration has been disabled`.

I tried to enable registration but it didn't help me at all. When registration is enabled I get the registration form after I click an activation link in MailHog.

What can be the reason of this problem? Is there any chance to activate user's account manually?

BronsonQuick:
Fixes #615.

### DIFF
--- a/puppet/modules/chassis/templates/local-config-db.php.erb
+++ b/puppet/modules/chassis/templates/local-config-db.php.erb
@@ -13,17 +13,17 @@ defined( 'ABSPATH' ) or define( 'ABSPATH', '<%= @wpdir %>/' );
 defined( 'WP_CONTENT_DIR' ) or define( 'WP_CONTENT_DIR', '<%= @contentdir %>' );
 
 <% if @network != false %>
+define( 'MULTISITE', true );
 if ( ! defined( 'WP_INSTALLING' ) || ! WP_INSTALLING ) {
-	define( 'MULTISITE', true );
 <% if @subdomains != false -%>
-	define( 'SUBDOMAIN_INSTALL', true );
+    define( 'SUBDOMAIN_INSTALL', true );
 <% else -%>
-	define( 'SUBDOMAIN_INSTALL', false );
+    define( 'SUBDOMAIN_INSTALL', false );
 <% end -%>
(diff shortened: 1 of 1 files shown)


rmccue:
This was intentionally set this way as otherwise installing via wp-cli in Puppet broke, I think. Not certain though.

BronsonQuick:
@rmccue Yeah I couldn't 100% remember the reason for the workaround you did but I remember you doing it for a reason and a workaround. I'm wondering if it might've even been a core multisite issue that's now been resolved upstream!

BronsonQuick:
I've tested this with subfolders, subdomains and single sites so this should be good to go.

puppet/modules/chassis/templates/local-config-db.php.erb
```diff
@@ -13,17 +13,17 @@ defined( 'ABSPATH' ) or define( 'ABSPATH', '<%= @wpdir %>/' );
 defined( 'WP_CONTENT_DIR' ) or define( 'WP_CONTENT_DIR', '<%= @contentdir %>' );
 
 <% if @network != false %>
+define( 'MULTISITE', true );
```
stuartshields: @BronsonQuick So adding MULTISITE inside the if statement will cause user activation links to not work, so we'd need to test this before we merge it in.

BronsonQuick: @stuartshields See: https://github.com/Chassis/Chassis/pull/629#issuecomment-494385875 currently they don't work now which is the issue I'm trying to fix.

Can you please try and test this too? Thanks dude!

stuartshields: Should really read stuff before commenting 👍

BronsonQuick:
I had a chat with @stuartshields and this is good to merge. Thanks Stu! 💖

Issue #628, opened 2019-05-17
Manually set machine name if the name is too generic


rmccue:
Vagrant sets the [name](https://www.vagrantup.com/docs/virtualbox/configuration.html#virtual-machine-name) of the machine to the directory name + timestamp by default. This isn't useful if the directory name is generic though, so we should detect this and set it manually.

IMO, we should set it to the parent directory's name if the directory name is one of:
* `chassis`
* `vagrant`

Any others? (Related: #404.)

roborourke:
Could possibly allow setting the name via the `config.yaml` too.

Rather than assume the parent directory I think it'd be better to look at the `paths.base` value and take that directory name. If it's still `chassis` or `vagrant` then underscore separate the absolute path on disk? Eg. `/Users/bob/chassis` becomes `users_bob_chassis`.

Maybe even just do the latter.

BronsonQuick:
After some messing around in a multi VM situation we need to do `config.vm.define` and `vb.name` will give it a name in the VirtualBox GUI. I did a quick and dodgy start on this and I pushed some work in progress so if either of you wanna pick it up then feel free to. I might get back to it on the weekend.

rmccue:
> Rather than assume the parent directory I think it'd be better to look at the `paths.base` value and take that directory name. If it's still `chassis` or `vagrant` then underscore separate the absolute path on disk? Eg. `/Users/bob/chassis` becomes `users_bob_chassis`.

I think that's going to be quite long; I'd have `users_rmccue_work_human made_hmn.md_chassis-hmnlocal` for example. IMO, just the directory name should do.

roborourke:
Playing with this a bit there's a challenge in terms of backwards compatibility. It's a breaking change if the expected VM name suddenly changes so it might just be best to leave it as the default unless the `machine_name` is set in the config.

BronsonQuick:
Alternatively we change https://github.com/Chassis/Chassis/pull/633/files#diff-259fe82e12a866f01123927480c7851bR111 to be `machine_name: default` and add an example name in the comment on the line above?

roborourke:
Agreed - in terms of backwards compat we should only use something other than `default` if explicitly defined in the config.

roborourke:
Alrighty - done & documented!

BronsonQuick:
Oopphhh! Nice one @roborourke 🍆

PR #633, opened 2019-05-29
Allow setting the machine name

BronsonQuick:
This starts work on #628. It's gonna need some refactoring and also adding some support for Windows to but it's a bit of a start in case someone else wants to pick it up in the meantime.


puppet/chassis.rb
```diff
@@ -134,6 +134,20 @@ def self.normalize_config(config)
...
 				end
 			end
+
+			machine_name = '' unless config["machine_name"]
+
+			# If we don't have a machine name in the config generate one from the directory structure.
+			if machine_name == ''
+				pwd = Dir.pwd.downcase
+				name = pwd.gsub("/","_")
+				if name[0] == "/"
```
roborourke: Is this a weird side effect of how `gsub` works? Pretty annoying! Thanks for making a start on this btw

BronsonQuick: Nah not at all. I was just following your comment:
>/Users/bob/chassis becomes users_bob_chassis
So I downcased it and needed to remove the first `/`

puppet/chassis.rb
```diff
@@ -134,6 +134,15 @@ def self.normalize_config(config)
 					config["mapped_paths"][path] = config["mapped_paths"]["base"]
 				end
 			end
+
+			machine_name = '' unless config["machine_name"]
+
+			# If we don't have a machine name in the config generate one from the directory structure.
+			if machine_name == ''
+				pwd = Dir.pwd.downcase
+				# Replace all non word characters with underscores then trim underscores.
```
BronsonQuick: FYI then machine name can actually contain spaces i.e. `rob is a champion` is a valid name.

config.yaml
```diff
@@ -110,6 +110,11 @@ php: 7.3
 # Maximum file upload size. This will set post_max_size and upload_max_filesize in PHP and client_max_body_size in Nginx.
 upload_size: 1024M
 
+# Vagrant machine name
+# This can be any string and helps to identify a specific VM
+# NOTE: If changed the machine will need to be re-created and should ideally be destroyed before any change
+machine_name: default
```
rmccue: Will this make the machine name statically `default`? The actual default in Vagrant is `basename( directory ) + machine_name + timestamp + some_other_number`

roborourke: Let me double check real quick

roborourke: I thought it was a special value - added if statements to only apply the setting if it's not `default`

rmccue: 😗 👌

Issue #634, opened 2019-06-01
Update the PHP version to 7.3


BronsonQuick:
I was doing some updates upstream to https://github.com/Chassis/core-dev and was about to post an issue about the WordPress Site Health feature warning about Chassis being on PHP 7.2 and I decided to check https://wordpress.org/about/requirements/ and it's now 7.3 and not 7.2 so should we update that?

PR #635, opened 2019-06-01
Bump the PHP version

BronsonQuick:
This "fixes" #634. It might not be an "issue" as such but I know I've had a bunch of people DMing me about all the Site Health checks and thinking it's a Chassis issue when it's not...but it kinda is according to devs.

Site_Health_Status_9_Chassis_Site__WordPress_2019-06-01_19-54-03

Maybe I should make yet another extension called `Site Health` that devs can use that installs all the extensions to get the `Site Health` stuff to 100%? I'm open to any and all feedback. Power in numbers y'all! 💯

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -105,7 +105,7 @@ synced_folders:
 
 # PHP version
 # Values: 5.6, 7.0, 7.1, 7.2, 7.3 (or 5.6.30)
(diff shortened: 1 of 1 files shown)


BronsonQuick:
Thanks both! 🤗

PR #638, opened 2019-06-06
Pass synced folders to guest

### LINKED ISSUE #637: Provide path from host to VM
For various dev tools (including [Query Monitor](https://querymonitor.com/blog/2019/02/clickable-stack-traces-and-function-names-in-query-monitor/)), it's useful to know what the path on the host is. VirtualBox doesn't provide this by default because it could be a security problem with some guests, but that doesn't really apply in our case.

VirtualBox provides the ability to pass data between the host and guest via [guest properties](https://www.virtualbox.org/manual/ch08.html#vboxmanage-guestproperty), so we can set these in our Vagrantfile.

We could also bring these in as PHP constants automatically, but probably not worth it.

rmccue:
Creates a new `/etc/chassis-constants` JSON file (world-readable) which contains a `synced_folders` object with the host's paths.

Fixes #637.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -68,6 +68,14 @@ Vagrant.configure("2") do |config|
 			vb.memory = CONF['virtualbox']['memory'] if CONF['virtualbox']['memory']
 			vb.cpus = CONF['virtualbox']['cpus'] if CONF['virtualbox']['cpus']
 		end
+
+		# Pass synched folders to guest
+		full_synched = {}
+		synced_folders.each do |from, to|
+			full_from = File.realpath from, base_path.to_s
+			full_synched[ to ] = full_from
+		end
+		vb.customize [ "guestproperty", "set", :id, "/Chassis/synced_folders", JSON.dump( full_synched ), "--flags", "TRANSIENT,RDONLYGUEST" ]
 	end
 
 	# We <3 Ubuntu LTS
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -49,6 +49,14 @@ if use_global_ext
 end
 
 Vagrant.configure("2") do |config|
+	# Set up synced folders.
+	synced_folders = CONF["synced_folders"].clone
+	synced_folders["."] = "/vagrant"
+
+	if use_global_ext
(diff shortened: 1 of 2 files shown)


rmccue:
As an example of integrating with QM (and VS Code):

```php
$data = json_decode( file_get_contents( '/tmp/chassis-constants' ) );

add_filter( 'qm/output/file_path_map', function ( $map ) use ( $data ) {
	$map = array_merge( $map, (array) $data->synced_folders );
	return $map;
} );
add_filter( 'qm/output/file_link_format', function( $format ) {
	return 'vscode://file/%f:%l';
} );
```

Vagrantfile
```diff
@@ -60,6 +68,14 @@ Vagrant.configure("2") do |config|
...
 			vb.cpus = CONF['virtualbox']['cpus'] if CONF['virtualbox']['cpus']
 		end
+
+		# Pass synced folders to guest
+		full_synced = {}
+		synced_folders.each do |from, to|
+			full_from = File.realpath from, base_path.to_s
+			full_synced[ to ] = full_from
+		end
+		vb.customize [ "guestproperty", "set", :id, "/Chassis/synced_folders", JSON.dump( full_synced ), "--flags", "TRANSIENT,RDONLYGUEST" ]
```
BronsonQuick: 🔢Any particular reason why you uppercased `/Chassis`?

BronsonQuick:
I'm thinking it might be worth adding:
```
$data = json_decode( file_get_contents( '/tmp/chassis-constants' ) );

add_filter( 'qm/output/file_path_map', function ( $map ) use ( $data ) {
	$map = array_merge( $map, (array) $data->synced_folders );
	return $map;
} );
```
into the code and:
```
add_filter( 'qm/output/file_link_format', function ( $format ) {
	return 'vscode://file/%f:%l';
} );
```
into the readme of https://github.com/Chassis/Query_Monitor

rmccue:
> 🔢Any particular reason why you uppercased `/Chassis`?

Yeah, it matches the guest properties built into VB, which are prefixed with `/VirtualBox/...`

Also will PR the changes to the QM extension 👍

PR #642, opened 2019-06-13
Increase max upload size

### LINKED ISSUE #640: Increase max upload size, or sync it with virtualbox.memory
The intention of the `post_max_size` and `upload_max_filesize` directives in PHP is to prevent giant files from being uploaded and eating up all the server's resources. For a local development environment, it's safe to jack up this limit as long as the VM can handle it.

Chassis currently sets a limit of 100MB for `upload_max_filesize`, `post_max_size`, and the corresponding `client_max_body_size` directive for Nginx. The `memory_limit` directive appears to be set to 128M, although my local sites have it set to 1024M and I can't remember if I changed that manually.

I think these default limits should be increased to 1GB all round. We've found that the kind of sites that we're building are using infrastructure (and processes) that support much higher limits, and upping these limits in Chassis for testing is quite a pain, especially for inexperienced users.

Alternatively, these limits could be dynamically populated based on the value of `virtualbox.memory` in config.yml.

The `Chassis/phpini` extension exists, but this doesn't fix the `client_max_body_size` directive for Nginx.

Thoughts?

BronsonQuick:
Fixes #640.


johnbillion:
Nice one. I _think_ this needs to affect `memory_limit` too because AFAIK PHP doesn't stream files during an upload, so the memory limit needs to be at least the same as the upload size limits. Needs verifying.

BronsonQuick:
@johnbillion I was actually wondering about that as well. I went with the assumption that it would stream. @rmccue I don't suppose you know that off the top of your head by any chance? #lazyweb 😁

BronsonQuick:
https://github.com/Chassis/Chassis/issues/640#issuecomment-501924237

config.yaml
```diff
@@ -107,6 +107,9 @@ synced_folders:
 # Values: 5.6, 7.0, 7.1, 7.2, 7.3 (or 5.6.30)
 php: 7.3
 
+# Maximum file upload size. This will set post_max_size and upload_max_filesize in PHP and client_max_body_size in Nginx.
+upload_size: 1024M
```
rmccue: I wonder whether we should have done
```yaml
php:
    version: 7.3
    upload_size: 1024M
```

instead. Pain to change now, but in hindsight...

BronsonQuick: I thought about that but it would mean I'd have to update all the extensions for that and that would take ages!

PR #651, opened 2019-07-18
Update our config

### LINKED ISSUE #650: Adjust our docs now have html extensions due to the latest update
@goldenapples noted that with the new Read The Docs config the URLs have changed so we should fix that up to avoid broken links from other projects.

I think we can fix this by adding:
```
builder: htmldir
```
to the new `.readthedocs.yaml` file then alter our Read The Docs admin settings back to `Sphinx HtmlDir`

BronsonQuick:
Fixes #650.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,6 +8,7 @@ version: 2
 # Build documentation in the docs/ directory with Sphinx
 sphinx:
   configuration: docs/conf.py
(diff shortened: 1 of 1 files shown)


BronsonQuick:
We'll need to alter our Read The Docs admin settings back to Sphinx HtmlDir once this is merged.

PR #658, opened 2019-08-05
Prepare for the extensions for Bionic Beaver

### LINKED ISSUE #657: Prepare extensions for Bionic Beaver updates
The newer version of Puppet doesn't allow classes to have [hyphens](https://coderwall.com/p/fvqsfw/avoid-hyphens-in-puppet-class-names) as it treats them as `minus` when it parses the manifests.

To get around this we should add in a `version: 2.1` to Chassis core then add in some logic to delete old extensions that contain a hyphen and upgrade them to underscores.

I think if we add that version and take code from here: https://github.com/Chassis/Chassis/pull/532/files#diff-34c580b66e353a044fcf7eee62a6b191R179-R195 we should be able to handle that. Once we have a PR for that merged, we can update the extensions that currently contain a hyphen with underscores and we'll be prepared for the Bionic Beaver update which we'll call version 3.

Old yaml files that have say `query-monitor` etc will get a redirect to `query_monitor` from Github so we should be fully backwards compatible.

BronsonQuick:
Fixes #657.

I've tested this with https://github.com/Chassis/Query_Monitor as it's the only extension I've prepared for Bionic Beaver.

A current installation with `query-monitor` installed in `extensions/query-monitor` (or has `- query-monitor` in a yaml file) remains untouched.

When I add `version: 3.0` to a custom yaml file the `extensions/query-monitor` extension is remove and cloned recursively into `extensions/query_monitor` as expected.

Once this is merged we will need to update the extensions that contain underscores in the repos and classes. I'll open issues upstream to in those extensions.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -8,6 +8,9 @@
 # - project/config.local.yaml          (global overrides)
 # - project/config.yaml                 (global defaults)
 
+# Add a version for Chassis for when we upgrade Ubuntu distributions. We're currently on version 2.1: https://github.com/Chassis/Chassis/releases/tag/2.1
+version: 2.1
+
 # Domain names to use for Vagrant
 #
 # The first name will be used as the machine's name, subsequent names will be
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -177,24 +177,31 @@ module Chassis
 			repo = extension
 		end
 
+		renamed_extensions = Array.new
 		# We need to add a version to allow for a seamless upgrade to Bionic Beaver.
 		if config["version"] >= 3
 			# Check for existing extensions that have a hyphen and remove them.
 			old_extensions = Dir.glob(@@dir + '/extensions/*')
 			old_extensions.each { |extension|
-				next unless extension.include? "-"
-					# Delete the old extension if it has a hyphen in it.
-					FileUtils.remove_dir(extension)
-					new_extension = extension.gsub(/-/, '_')
-					puts "We've renamed #{extension} to #{new_extension} as hyphens aren't allowed in Puppet: https://puppet.com/docs/puppet/5.4/lang_reserved.html#classes-and-defined-resource-types"
-					puts "If you have #{extension} in one of your yaml configuration files we would recommend altering it to #{new_extension}."
-				end
-		    }
+			next unless extension.include? "-"
+				# Delete the old extension if it has a hyphen in it.
+				FileUtils.remove_dir(extension)
+				new_extension = extension.gsub(/-/, '_')
+				renamed_extensions << "- #{extension} to #{new_extension}"
+			}
 		    # Puppet have taken a hard stance on not allowing hyphens in class names.
 		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/\.git$/, '').gsub(/-/, '_').downcase
 		else
 		    # Leave extensions as they were for Xenial and below for backwards compatibility.
 		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/\.git$/, '').downcase
+		end
+
+		if ! renamed_extensions.empty?
+			puts("\e[33mWe're renaming the following extensions:")
+			renamed_extensions.each {|extension|
+				puts extension
+			}
+			puts("\e[0m")
 		end
 
 		system("git clone %s %s --recursive" % [repo, Shellwords.escape(folder)] ) unless File.exist?( folder )
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -187,6 +187,7 @@ module Chassis
 					FileUtils.remove_dir(extension)
 					new_extension = extension.gsub(/-/, '_')
 					puts "We've renamed #{extension} to #{new_extension} as hyphens aren't allowed in Puppet: https://puppet.com/docs/puppet/5.4/lang_reserved.html#classes-and-defined-resource-types"
+					puts "If you have #{extension} in one of your yaml configuration files we would recommend altering it to #{new_extension}."
 				end
 		    }
 		    # Puppet have taken a hard stance on not allowing hyphens in class names.
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -186,7 +186,7 @@ module Chassis
 					# Delete the old extension if it has a hyphen in it.
 					FileUtils.remove_dir(extension)
 					new_extension = extension.gsub(/-/, '_')
-					puts "We've upgraded #{extension} to #{new_extension}"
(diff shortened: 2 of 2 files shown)


config.yaml
```diff
@@ -8,6 +8,9 @@
 # - project/config.local.yaml          (global overrides)
 # - project/config.yaml                 (global defaults)
 
+# Add a version for Chassis for when we upgrade Ubuntu distributions. We're currently on version 2.1: https://github.com/Chassis/Chassis/releases/tag/2.1
+version: 2.1
```
rmccue: Should this be split out? 🤔

puppet/chassis.rb
```diff
@@ -177,7 +177,24 @@ def self.install_extension(extension)
...
+		# We need to add a version to allow for a seamless upgrade to Bionic Beaver.
+		if config["version"] >= 3
+			# Check for existing extensions that have a hyphen and remove them.
+			old_extensions = Dir.glob(@@dir + '/extensions/*')
+			old_extensions.each { |extension|
+				if extension.include? "-"
+					# Delete the old extension if it has a hyphen in it.
+					FileUtils.remove_dir(extension)
+					new_extension = extension.gsub(/-/, '_')
+					puts "We've upgraded #{extension} to #{new_extension}"
```
rmccue: "renamed" might be better to be more explicit, plus maybe we should link out somewhere to explain why we're doing it? (e.g. link to this issue/PR)

puppet/chassis.rb
```diff
@@ -177,7 +177,24 @@ def self.install_extension(extension)
 			repo = extension
 		end
 
-		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '').downcase
+		# We need to add a version to allow for a seamless upgrade to Bionic Beaver.
+		if config["version"] >= 3
+			# Check for existing extensions that have a hyphen and remove them.
+			old_extensions = Dir.glob(@@dir + '/extensions/*')
+			old_extensions.each { |extension|
+				if extension.include? "-"
```
rmccue: Could we `continue unless extension.include? "-"`?

puppet/chassis.rb
```diff
@@ -177,7 +177,24 @@ def self.install_extension(extension)
...
+					# Delete the old extension if it has a hyphen in it.
+					FileUtils.remove_dir(extension)
+					new_extension = extension.gsub(/-/, '_')
+					puts "We've upgraded #{extension} to #{new_extension}"
+				end
+		    }
+		    # Puppet have taken a hard stance on not allowing hyphens in class names.
+		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '').gsub(/-/, '_').downcase
```
rmccue:
```suggestion
		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/\.git$/, '').gsub(/-/, '_').downcase
```

puppet/chassis.rb
```diff
@@ -177,7 +177,24 @@ def self.install_extension(extension)
...
+		else
+		    # Leave extensions as they were for Xenial and below for backwards compatibility.
+		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '').downcase
```
rmccue:
```suggestion
		    folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/\.git$/, '').downcase
```

config.yaml
BronsonQuick: I thought about that but seeing we've been slack with version I figured we can slip it in. 🍆

puppet/chassis.rb
BronsonQuick: Ahh good thinking. In Ruby `continue` is `next` so I've added that.

puppet/chassis.rb
BronsonQuick: Good call. Done!

puppet/chassis.rb
```diff
@@ -177,7 +177,25 @@ def self.install_extension(extension)
...
+		if config["version"] >= 3
+			# Check for existing extensions that have a hyphen and remove them.
+			old_extensions = Dir.glob(@@dir + '/extensions/*')
+			old_extensions.each { |extension|
+				next unless extension.include? "-"
+					# Delete the old extension if it has a hyphen in it.
+					FileUtils.remove_dir(extension)
+					new_extension = extension.gsub(/-/, '_')
+					puts "We've renamed #{extension} to #{new_extension} as hyphens aren't allowed in Puppet: https://puppet.com/docs/puppet/5.4/lang_reserved.html#classes-and-defined-resource-types"
+					puts "If you have #{extension} in one of your yaml configuration files we would recommend altering it to #{new_extension}."
```
rmccue: Might want to move this outside of the loop so it's only done once? Also, I'd say we can just stick to the first sentence.

Can we handle the config bit ourselves in the config normalisation step? Maybe warn there instead.

puppet/chassis.rb
```diff
@@ -177,7 +177,25 @@ def self.install_extension(extension)
 			repo = extension
 		end
 
-		folder = @@dir + '/extensions/' + extension.split('/').last.gsub(/.git$/, '').downcase
+		# We need to add a version to allow for a seamless upgrade to Bionic Beaver.
+		if config["version"] >= 3
+			# Check for existing extensions that have a hyphen and remove them.
+			old_extensions = Dir.glob(@@dir + '/extensions/*')
+			old_extensions.each { |extension|
+				next unless extension.include? "-"
```
rmccue: This isn't a block now, so I think we can deindent the next lines (and remove `end`)?

config.yaml
```diff
@@ -8,6 +8,9 @@
 # - project/config.local.yaml          (global overrides)
 # - project/config.yaml                 (global defaults)
 
+# Add a version for Chassis for when we upgrade Ubuntu distributions. We're currently on version 2.1: https://github.com/Chassis/Chassis/releases/tag/2.1
```
rmccue:
```suggestion
# Add a version for Chassis for when we upgrade Ubuntu distributions.
# We're currently on version 2.1: https://github.com/Chassis/Chassis/releases/tag/2.1
```

puppet/chassis.rb
BronsonQuick: Done and done!

puppet/chassis.rb
BronsonQuick: Done!

BronsonQuick:
@rmccue Thanks for the re-review. I've added those updates. Here's what the warnings look like now:

renaming
yaml

BronsonQuick:
@rmccue Ping for another review please!

PR #659, opened 2019-08-06
Set nginx fastcgi_buffer config to stop 502 errors

jenkoian:
A lot of my admin ajax scripts (such as adding a custom attribute in woocommerce) were giving a 502 error. Having dug into this issue I came across https://stackoverflow.com/a/23845727/380054 which fixed the issue for me.

If we do decide to go with this we may want to have a read of: https://gist.github.com/magnetikonline/11312172#determine-fastcgi-response-sizes it seems that a one size fits all solution may not be the best option, however sensible defaults with documentation on how to change these values to suit is perhaps better than broken environments.

### DIFF
--- a/puppet/modules/chassis/templates/site.nginx.conf.erb
+++ b/puppet/modules/chassis/templates/site.nginx.conf.erb
@@ -8,6 +8,8 @@ server {
 	fastcgi_split_path_info ^(.+\.php)(/.+)$;
 	fastcgi_index index.php;
 	fastcgi_read_timeout 900;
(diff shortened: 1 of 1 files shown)


johnbillion:
+1, I'm generally in favour of this, although:

1. It could mask a problem that you won't encounter until you hit production and it turns out your production box also has the default proxy size limit in place. Not sure if anything can be done about that, though.
2. You might need to change more than just those two directives, see comments in this thread: https://wordpress.org/support/topic/502-bad-gateway-36/

PR #663, opened 2019-08-10
Fix nginx logging

### LINKED ISSUE #662: Access logs in nginx aren't being logged
@pdewouters pointed out that access logs for nginx weren't being synced by Vagrant into `logs/nginx/access.log`.

I sshed into a new Chassis box, browsed a few URLs, SSHed into the VM and did `cat /var/log/nginx/access.log` and noticed we didn't have any access logs there either.

This is a file permissions issue as nginx runs as the `www-data` user so we should set the synced to run as this user as well.

BronsonQuick:
Fixes #662. 

`nginx -t` showed the warning:
>[warn] the "user" directive makes sense only if the master process runs with super-user privileges, ignored in /etc/nginx/nginx.conf:1

So I removed that as well.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -166,7 +166,7 @@ Vagrant.configure("2") do |config|
 	mount_opts = CONF['nfs'] ? [] : ["dmode=777","fmode=777"]
 
 	synced_folders.each do |from, to|
-		config.vm.synced_folder from, to, :mount_options => mount_opts, :nfs => CONF['nfs']
+		config.vm.synced_folder from, to, :mount_options => mount_opts, :nfs => CONF['nfs'], :group => 'www-data', :owner => 'www-data'
 
 		# Automatically use bindfs if we can.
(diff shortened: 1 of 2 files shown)


peterwilsoncc:
Before approving, I’m trying to confirm the Nginx service doesn’t get started by root but unable to do so. If it does get started by root then the user line makes sense & the warning is a false negative 

Vagrant config changes make sense.

BronsonQuick:
The ngnix worker runs as www-data. I can't remember the command I ran which listed the master and worker processes

BronsonQuick:
@peterwilsoncc Here's a screenshot of `systemctl status nginx`

chassis__vagrantvagrant___ssh__vagrant_ssh__15841_2019-08-16_09-51-24

pdewouters:
works for me thanks!

BronsonQuick:
@pdewouters Awesome! Thanks again for letting me know about it!

PR #666, opened 2019-08-19
Warn when .local isn't used

### LINKED ISSUE #571: Warn the user if they use a domain other than `.local`
As a user I should get a warning if I've tried to use a domain other than `.local`

BronsonQuick:
Fixes #571.


BronsonQuick:
extensions__fish_VolumesSiteschassisextensions__-fish__15841_2019-08-19_16-52-35

PR #668, opened 2019-08-19
Add details about ngrok version 2.2.8

### LINKED ISSUE #644: Vagrant Share doesn't work with ngrok after version 2.2.8
@kadamwhite was unable to get ngrok to work on version `2.3.29`. After some debugging it seems that something is broken in ngrok so for the moment we should document that version `2.2.8` is required to work.

The ngrok archives are available here: https://dl.equinox.io/ngrok/ngrok/stable/archive

BronsonQuick:
Fixes #644.

### DIFF
--- a/docs/guides.rst
+++ b/docs/guides.rst
@@ -93,11 +93,11 @@ Vagrant Share enables the ability to generate a temporary URL which you can shar
 
    Run the following command in a terminal to install the `Vagrant Share plugin`_. ``vagrant plugin install vagrant-share``.
 
-2. **Install ngrok**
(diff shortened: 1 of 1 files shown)


docs/guides.rst
```diff
@@ -109,11 +109,14 @@ Vagrant Share enables the ability to generate a temporary URL which you can shar
 
    Navigate to the URL that ngrok generated.
 
+**Note**: ngrok Version 2.2.8 is required due to this known `bug`_:
```
rmccue: btw there's a trailing `:` here :)

PR #671, opened 2019-08-20
Make sure we update apt

BronsonQuick:
This will make sure the PPA updates which will fix the PHP issue in #582 but it doesn't add the weekly PPA updates so we'll leave that issue open for now.


BronsonQuick:
Oopss. I should've tested this. This adds a dependency cycle. 🙈

PR #673, opened 2019-08-20
Remove extra_hosts for single sites

### LINKED ISSUE #672: `conflicting server name "vagrant.local"`
On a single site `/logs/nginx/error.log` shows the follows warnings:
```
2019/08/20 07:46:08 [warn] 3574#3574: conflicting server name "vagrant.local" on 0.0.0.0:80, ignored
2019/08/20 07:46:08 [warn] 3574#3574: conflicting server name "vagrant.local" on [::]:80, ignored
```
This is because of this line: https://github.com/Chassis/Chassis/blob/master/puppet/modules/chassis/manifests/site.pp#L18

BronsonQuick:
Fixes #672. 

This changes the `server` block in Nginx from:
`server_name vagrant.local vagrant.local *.ngrok.io;`
to
`server_name vagrant.local *.ngrok.io;`

### DIFF
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -15,7 +15,7 @@ define chassis::site (
 	$sitename       = 'Chassis Site',
 ) {
 	$extra_hosts = join($hosts, ' ')
-	$server_name = rstrip("${name}")
+	$server_name = rstrip($name)
 	file { $wpdir:
 		ensure => directory
 	}
--- a/puppet/modules/chassis/manifests/site.pp
+++ b/puppet/modules/chassis/manifests/site.pp
@@ -15,7 +15,7 @@ define chassis::site (
 	$sitename       = 'Chassis Site',
 ) {
 	$extra_hosts = join($hosts, ' ')
-	$server_name = rstrip("${name} ${extra_hosts}")
+	$server_name = rstrip("${name}")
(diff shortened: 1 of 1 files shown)


puppet/modules/chassis/manifests/site.pp
```diff
@@ -15,7 +15,7 @@
 	$sitename       = 'Chassis Site',
 ) {
 	$extra_hosts = join($hosts, ' ')
-	$server_name = rstrip("${name} ${extra_hosts}")
+	$server_name = rstrip($name)
```
ntwb:
```suggestion
	$server_name = rstrip(${name})
```

Should this remain in curly brackets?

Also what about the hosts array of hosts, is this affected by this change?

Also, what about these instances?
https://github.com/Chassis/Chassis/blob/5f8efbae510a931365f5294bcddca2ea7a4715db/puppet/modules/chassis/manifests/network.pp#L19

https://github.com/Chassis/Chassis/blob/5f8efbae510a931365f5294bcddca2ea7a4715db/puppet/modules/chassis/manifests/network.pp#L22

BronsonQuick: >Should this remain in curly brackets?
Negative.

>Also what about the hosts array of hosts, is this affected by this change?
There should only be more than one host if they are using multisite.

>Also, what about these instances?
They are fine as they are for multisites.

PR #690, opened 2019-08-26
Exclude parent directory from hyphenated class check

### LINKED ISSUE #689: Parent dir with - in name triggers extension renaming notice
The check to display the "we are renaming your extensions" warning appears to trigger based on the presence of a `-` dash character anywhere within the directory path, where it really only matters if the dash occurs within the `extensions` folder tree. This is most obvious when using an extension with no dash at all, such as `xdebug`, within a Chassis root folder named with a hyphen in the folder title.

For example, I have Chassis checked out to `/wp/core-vm` for my WP development repository. When I provision, I see this error:
```
We're renaming the following extensions:
- /wp/core-vm/extensions/xdebug to /wp/core_vm/extensions/xdebug
```
No rename has occurred, nor was one needed. The `core-vm` directory part is outside of the virtual machine's awareness so it will not impact provisioning.

I believe we could improve the logic for displaying this error by excluding the Chassis root path before detecting for dashes.

System info, though I don't believe it's relevant this time:

Ubuntu 19.04, Vagrant 2.2.4, VBox 6.0.6. Custom config file specifying `chassis/xdebug` in my extensions list.

kadamwhite:
A parent directory (Chassis checkout repo root) such as `.../my-vm` has no impact on Puppet's new stricter rules around using `-` instead of `-` in class names, but if Chassis was checked out into a hyphenated directory, our logic previously detected and warned on the "-" in that parent directory even if no hyphenated Puppet class was present.

If we do not exclude the parent directory from the string when testing for or, more critically, creating new correct path strings, we end up trying to clone the repositories into a non-existent directory path.

Fixes #689

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -197,17 +197,19 @@ module Chassis
 			# Check for existing extensions that have a hyphen and remove them.
 			old_extensions = Dir.glob(@@dir + '/extensions/*')
 			old_extensions.each { |extension|
-			next unless extension.include? "-"
+			# Remove @@dir before "-" include check in case of hyphenated parent directories.
+			next unless extension.gsub(@@dir, '').include? "-"
 				# Delete the old extension if it has a hyphen in it.
 				FileUtils.remove_dir(extension)
-				new_extension = extension.gsub(/-/, '_')
+				# Swap "-" for "_" in the non-parent-directory part of the path.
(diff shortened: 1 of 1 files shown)


BronsonQuick:
I can confirm this works a treat. Thanks again @kadamwhite 💖

PR #703, opened 2019-09-11
Allow extensions to pass in configuration values

### LINKED ISSUE #702: Allow extensions to declare config values
Currently we have the ability to specify `dependencies` in extensions through `chassis.yaml` but we should also be able to add other configuration values as well so we can specify version numbers and other variables.

BronsonQuick:
Fixes #702.

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -76,6 +76,14 @@ module Chassis
 			end
 		end
 
+		# Get the extension configurations.
+		ext_config = self.load_extension_config(_config)
+
+		# Merge the extension configuration into our main configuration.
+		if ext_config.is_a?(Hash)
+			_config.merge!(ext_config)
+		end
+
 		unless _config["database"].has_key? "prefix"
 			puts "ERROR: database.prefix is required as of 2016-11-25 but was not found."
 			raise "Could not load Chassis configuration"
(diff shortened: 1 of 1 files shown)


puppet/chassis.rb
```diff
@@ -76,6 +76,14 @@ def self.load_config()
 			end
 		end
 
+		# Get the extension configurations.
+		ext_config = self.load_extension_config(_config)
+
+		# Merge the extension configuration into our main configuration.
+		if ext_config.is_a?(Hash)
```
roborourke: I'm not sure an extension config should be able to override your local configs. Is this more of a way to allow some defaults and expose values to other extensions rather than setting hard requirements?

BronsonQuick: Yeah it's more to add the ability expose new values to the extensions...although, not that I think about it, being able to override some defaults in an extension could come in handy further down the track #undocumentedfeature

roborourke: I think it'd be super confusing potentially. I could define the PHP version for example, and despite the setting I have in my local config it'd be ignored. How would you debug that?

rmccue: They should be inserted into the override chain before `config.yaml`, but after all the local configs.

BronsonQuick:
@roborourke @rmccue Alrighty, based on a combo of your feedback I've refactored this now so the config order is:

`chassis.yaml` from the extensions can be overridden by `config.yaml` in Chassis core then `config.local.yaml` then `content/config.yaml` then `content/config.local.yaml`

PR #711, opened 2019-09-16
Add MIT license.

### LINKED ISSUE #708: Add GPL licence

BronsonQuick:
Fixes #708.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 Chassis
+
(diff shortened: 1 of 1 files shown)


rmccue:
Given Chassis includes WordPress, we might need to be GPL 🤔

BronsonQuick:
I wondered about that and thought they were compatible. I'll swap it to GPL.

rmccue:
Yeah, they are in one direction but not the other: GPL projects can include MIT, but not vice versa.

PR #721, opened 2019-10-09
Remove chassis-updated after provisioning

### LINKED ISSUE #720: Initial provisioning errors with basebox
We're hitting errors when provisioning now that we're using the basebox (#476). I think what's happened is that the URLs are expiring or something.

To replicate, simply start a new machine on the latest commit. Initial provisioning will fail with errors like:

```
==> default: Err:1 http://ppa.launchpad.net/ondrej/php/ubuntu xenial/main amd64 php7.2-common amd64 7.2.22-1+ubuntu16.04.1+deb.sury.org+1
==> default:   404  Not Found
==> default: Err:2 http://ppa.launchpad.net/ondrej/php/ubuntu xenial/main amd64 php7.2-zip amd64 7.2.22-1+ubuntu16.04.1+deb.sury.org+1
==> default:   404  Not Found
```

([Full log here](https://gist.github.com/abhishek-kaushik/ceaf855cd71108c95279efc01b6b342b).)

rmccue:
Fixes #720.

This could be a temporary fix if we can come up with a time-based solution (#582) instead.

### DIFF
--- a/puppet/preparebox.sh
+++ b/puppet/preparebox.sh
@@ -1,1 +1,4 @@
-# This is a placeholder file in case we need to run other scripts for the base box
+# Remove anything used during provisioning which needs to be re-done for each
+# individual box.
+
+rm /etc/chassis-updated


rmccue:
This will also necessitate rebuilding the base box and publishing a new version; I noticed (https://github.com/Chassis/Chassis/issues/720#issuecomment-540047297) that the current one might have some inadvertent changes from a `config.local.yaml` btw.

PR #732, opened 2019-11-15
Update apt weekly

### LINKED ISSUE #582: Update Apt automatically

BronsonQuick:
Fixes #582.

### DIFF
--- a/puppet/manifests/development.pp
+++ b/puppet/manifests/development.pp
@@ -18,7 +18,12 @@ if $loadable_extensions {
 		config => $config,
 	}
 }
-include apt
+
+class { 'apt':
+	update => {
+		frequency => 'weekly',
+	},
+}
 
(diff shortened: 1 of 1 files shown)


roborourke:
Hm, could've sworn I saw this change already. Maybe in another branch though. Why wouldn't we set this to `always`?

BronsonQuick:
@roborourke Yeah it was in there a while back but I reverted it as there was a conflict but the conflict was in an extensions.

Ray Ray said we should go with weekly when I talked to him about it ages ago so that was the reason for that.

PR #744, opened 2019-11-20
Remove the _global folder

### LINKED ISSUE #743: Make sure the `_global` folder isn't used in the extension update check.
Currently we get the following notice when an update check runs:
```
The following Chassis extension appears to be out of date: _global. This may cause provisioning to fail! Would you like to update them now? [Y/n]:
```

BronsonQuick:
Fixes #743.

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -322,6 +322,8 @@ module Chassis
 			puts "\033[A\e[32mChecking for Chassis extension updates...\e[0m"
 			extensions = Dir.glob(@@extension_dir + '/*').map { |directory| File.basename( directory ) }
 			updates = self.updates_check(extensions, @@extension_dir)
+			# Remove the _global folder as it's a synced folder.
+			updates.delete('_global')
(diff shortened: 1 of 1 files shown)


puppet/chassis.rb
```diff
@@ -322,6 +322,8 @@ def self.update_extensions(config)
 			puts "\033[A\e[32mChecking for Chassis extension updates...\e[0m"
 			extensions = Dir.glob(@@extension_dir + '/*').map { |directory| File.basename( directory ) }
 			updates = self.updates_check(extensions, @@extension_dir)
+			# Remove the _global folder as it's a synced folder.
+			updates.delete('_global')
```
rmccue: Presumably `.delete` means remove from the array? Surprised it's not `.delete!()` or similar given it's modifying the array.

rmccue: Yep, checked: https://ruby-doc.org/core-2.5.0/Array.html#method-i-delete - weird, seems inconsistent for Ruby!

PR #752, opened 2019-11-30
Fix global extensions properly

### LINKED ISSUE #737: Fix global extensions
Whilst looking into https://github.com/Chassis/Xdebug/issues/56#issuecomment-540989956 it looks like we've lost the ability to install global extensions somewhere along the line. I think we'll have to use `git bisect` to track down how we lost it.

BronsonQuick:
Fixes #737.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -16,7 +16,7 @@ configuration while still allowing overriding:
 * ``project/content/config.yaml`` - project-specific defaults
 * ``project/config.local.yaml`` - global overrides
 * ``project/config.yaml`` - global defaults
-* ``~/.chassis/config.yml`` - global update defaults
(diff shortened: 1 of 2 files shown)


BronsonQuick:
Once this is in then we can merge it into Bionic Beaver and we should be nearly ready to ship a Bionic Beaver base box finally! 😅

BronsonQuick:
@roborourke Nah. The directory.change of the global.config was just a mistake on my part and it wasn't in there long enough. I know it affected @ntwb briefly but this will fix it all up. Thanks for the speedy review! I'll do some testing of Bionic Beaver again in the next couple of days then after that I'll whip up the Vagrsnt plugin for the updates.

PR #766, opened 2019-12-30
Add box minimisation

### LINKED ISSUE #763: Minimise base box size
VVV has some scripts to minimise their base box size: e.g. https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/minimize.sh and https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/cleanup.sh. We might be able to use some of those to reduce the base box size for Chassis. We could also compare packages that are removed in Alpine Linux as well to see what else we could remove.

Props @tomjn for pointing out the minimize script.

BronsonQuick:
Fixes #763. 

This pretty much halves the base box size! It goes from `1.28G` to `652MB`. 😱🤘🕺

### DIFF
--- a/buildbox.sh
+++ b/buildbox.sh
@@ -35,10 +35,24 @@ echo "\033[0;32mThe WordPress database has been successfully dropped.\033[0m"
 
 echo "\033[0;32mWe are now ready to halt the VM and generate the base box.\033[0m"
 
+# Clean up the items on the box we don't need.
+
+echo "\033[0;32mInitialising box cleanup...\033[0m"
+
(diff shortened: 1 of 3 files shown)


tomjn:
@loreleiaurora someone applied your magic and got results :)

Originals:

 - https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/cleanup.sh
 - https://github.com/Varying-Vagrant-Vagrants/BoxBuilder/blob/master/ubuntu-18.04/scripts/minimize.sh

PR #769, opened 2020-01-08
Update base box

### LINKED ISSUE #765: Update the base box
We should build a new base box `3.0.1` because a new version of PHP 7.3 has shipped.
The base box will include:

- `software-properties-common` `0.96.24.32.11` to `0.96.24.32.12`
- PHP 7.3 `7.3.12` to `7.3.13`.
- Box minimisation.
- Duplicate mime type fix.

It would be good to look at https://github.com/Chassis/Chassis/issues/763 before we generate the base box too.

BronsonQuick:
Fixes #765.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -10,7 +10,7 @@
 
 # Add a version for Chassis for when we upgrade Ubuntu distributions.
 # We did start tagging releases a while back: https://github.com/Chassis/Chassis/releases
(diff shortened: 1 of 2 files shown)


BronsonQuick:
I've built the base box and it's been uploaded. I've drafted a release and I'll publish that once this is merged.

PR #772, opened 2020-02-03
Add monkey patch for host only issue

### LINKED ISSUE #771: Consider adding a monkey patch to core.
Yesterday I hit an issue on a friends computer where this error occured:
>A host only network interface you're attempting to configure via DHCP
already has a conflicting host only adapter with DHCP enabled. The
DHCP on this adapter is incompatible with the DHCP settings. Two
host only network interfaces are not allowed to overlap, and each
host only network interface can have only one DHCP server. Please
reconfigure your host only network or remove the virtual machine
using the other host only network.

I tried to work around it in several ways and the only way I could fix it was with this [monkey patch](https://github.com/hashicorp/vagrant/issues/8878#issuecomment-345112810).

I've had one other person report this issue as well so we might need to consider adding it to Chassis core.

BronsonQuick:
Fixes #771. See: https://github.com/hashicorp/vagrant/issues/8878#issuecomment-345112810

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -9,6 +9,13 @@ if Gem::Version.new(Vagrant::VERSION) < Gem::Version.new("1.5.0")
 	exit 1
 end
 
+# See https://github.com/hashicorp/vagrant/issues/8878 and https://github.com/Chassis/Chassis/issues/771
+class VagrantPlugins::ProviderVirtualBox::Action::Network
+	def dhcp_server_matches_config?(dhcp_server, config)
+		true
+	end
+end
+
 # Check that submodules have been loaded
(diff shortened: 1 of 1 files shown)


BronsonQuick:
@roborourke Thank you Sir Yeah I don't get the internals of it either but the patch works and stops breakage. I'm keeping an eye on the other ticket so hopefully we can rip it out when it's fixed!

mathiasaerts:
Could be related to hashicorp/vagrant#11403 for anyone using Vagrant 2.2.7 with Virtualbox 6.1. A fix has been merged to master and will be included in the next release.

BronsonQuick:
@mathiasaerts Champion. Thanks so much for linking that issue. 💖 I'll open a new issue to revert the monkey patch for when Vagrant `2.2.8` is released!

PR #776, opened 2020-03-04
Remove duplicate line.

jenkoian:


rmccue:
This line is actually intentionally in here, as WP can break in subtle ways if there's only one item in the array. We should clarify that with a comment probably.

johnbillion:
We have created a monster

jenkoian:
@rmccue thanks, will close. Yeah comment be useful 👍

PR #796, opened 2020-05-03
Focal Fossa

### LINKED ISSUE #787: Add option to allow MySQL server overrides
During a database import I was receiving errors along the following lines:
`Invalid default value for 'comment_date'` this was due to values like `0000-00-00 00:00:00` which means I needed to override `mysqld`'s defaults.

To do this we need to add `override_options` to `mysql::server` as an option so I can do something like the following:
```
	mysqld => {
		'sql_mode' => 'ERROR_FOR_DIVISION_BY_ZERO,NO_ZERO_DATE,NO_ZERO_IN_DATE,NO_AUTO_CREATE_USER',
	}
```
Let's pass in no options by default and allow an override in our configs.

### LINKED ISSUE #792: Update base box to Ubuntu 20.04 LTS Focal Fossa

BronsonQuick:
This adds support for Focal Fossa and fixes #792.

I've tested this with:
- PHP 5.6
- PHP 7.0
- PHP 7.1
- PHP 7.2
- PHP 7.3
- PHP 7.4

There are some warnings that are thrown such as:
`default: /usr/lib/ruby/vendor_ruby/puppet/util.rb:461: warning: URI.escape is obsolete` which has been reported in upstream in Puppet.

I've uploaded the new base box to Vagrant Cloud.

I also fixed #787 as well as I needed to do that to add some default character sets for MySQL.


puppet/manifests/development.pp
```diff
@@ -30,8 +30,27 @@
 	ensure => installed
 }
 
+$mysql_defaults = {
+	client => {
+		'default-character-set' => 'utf8'
```
roborourke, line 35: Should this be utf8mb4?

BronsonQuick: Herp, derp! Indeed it should. Fixed!

BronsonQuick:
I've updated the base box with `utf8mb4`

puppet/modules/chassis/manifests/hosts.pp
```diff
@@ -3,20 +3,14 @@
...
-		path    => '/bin:/usr/bin',
-		command => 'pip install --upgrade pip==9.0.3',
-		require => Package['python-pip']
-	}
-
 	ensure_packages( ['mdns-publisher'], {
 		ensure   => present,
-		provider => 'pip',
-		require  => [ Package['python-pip'], Package['libdbus-glib-1-dev'], Exec['upgrade pip'] ],
+		provider => 'pip3',
```
roborourke, line 12: Can't think off the top of my head but cold there be any other extensions affected by updating to python 3? I didn't find anything on a search of the chassis org with a provider of 'pip' so I think we're ok.

BronsonQuick: Yeah this is the only python dependency in Chassis and the extensions so I think we should be good. I will do due diligence and test it with the majority of the extensions just to be safe. Thanks for the speedy review 🥰🍆 I didn't want this one sitting in the queue for months like Bionic Beaver!

BronsonQuick:
It's been a while since I've checked this out again and this is still throwing `/usr/lib/ruby/vendor_ruby/puppet/util.rb:461: warning: URI.escape is obsolete` warnings 😢

tomjn:
@BronsonQuick what's the blockers around this at the moment?

BronsonQuick:
@tomjn It's the massive amount of warnings that puppet throws. I've been able to successfully provision it since https://github.com/Chassis/Chassis/pull/796#issuecomment-709016964 but the warnings are crazy. I tried a few things to suppress them but didn't have any luck.

BronsonQuick:
@tomjn Here's an example of some of the output:
chassis__fish_VolumesSiteschassis__-fish__19056_2021-12-06_10-25-41

If you want to replicate it then check out this branch, edit `config.yaml` so it has `_mode: base` at the bottom and run a `vagrant up`. Note that you won't see the normal Chassis message showing you the WP URL, admin username and password but if you got to `http://vagrant.local` you'll see your WP site.

BronsonQuick:
@tomjn Thanks for the ping on this! I just had another few cracks at suppressing the warnings in various ways this afternoon and I've finally managed to do it with: [`4f477ea` (#796)](https://github.com/Chassis/Chassis/pull/796/commits/4f477ea9f38c362c6bf43149b625192d230f8277)
This has been bugging me and on my mind for over a year 😅 

chassis__fish_VolumesSiteschassis__-fish__12557_2021-12-06_18-59-56

I'm going to now test against the following versions of PHP:
- [ ] 5.6
- [ ] 7.0
- [ ] 7.1
- [ ] 7.2
- [ ] 7.3
- [ ] 7.4
- [ ] 8.0
- [ ] 8.1

I'll tick them off as I go and if all goes well I can finally build and ship a new Chassis base box on Focal Fossa! 🤞🏻

tomjn (approved):

BronsonQuick:
@tomjn Thanks for the approval. I've just checked all versions of PHP and found a little bug for PHP 8.1 so I've fixed that in `main` and in this branch too. I'm about to build a new base box, upload it, check it once more then start writing some release notes! I should be done in about an hours time as I've gotta get some dinner into me as well.

BronsonQuick:
I've created base box 4.0.0, uploaded to Vagrant Cloud, tested it and it's all good so I'll merge this now then publish the release.

PR #805, opened 2020-06-06
Dynamically generate wp-cli.yml

### LINKED ISSUE #804: Make the machine name dynamic for WP-CLI
If I have set the machine name in Chassis then the `@local` alias for `wp-cli.yml` is incorrect.

If I change the machine name in a configuration file. e.g.
`machine_name: yolo` and run `wp @local --info` I will see the following error:

```
The machine with the name 'default' was not found configured for this Vagrant environment.
```
To avoid this we should set the machine name dynamically. So if I set my machine name to `machine_name: yolo` we should generate a `wp-cli.yml` as such:
```
@local:
  ssh: vagrant:yolo
```

BronsonQuick:
Fixes #804.


puppet/modules/chassis/manifests/wp.pp
```diff
@@ -83,4 +84,9 @@
 	file { '/home/vagrant/.wp-cli/config.yml':
 		content => template('chassis/wp-cli.yml.erb')
 	}
+
+	file { '/vagrant/wp-cli.yml':
+		content => template('chassis/wp-cli.local.yml.erb')
```
rmccue, line 89: `wp-cli.local.yml` is a real thing (for overrides), so imo let's call this `wp-cli.yml.erb`, and rename the old one to `wp-cli-config.yml.erb`

puppet/modules/chassis/templates/wp-cli.local.yml.erb
```diff
@@ -0,0 +1,3 @@
+# This is generated automatically by Puppet. Edit at your own risk.
```
rmccue, line 1: Should note:
```suggestion
# This is generated automatically by Puppet. Rather than editing,
# override these settings in a wp-cli.local.yml file.
# See https://make.wordpress.org/cli/handbook/references/config/
```

puppet/modules/chassis/manifests/wp.pp
BronsonQuick: Awesome. Thanks for that. I originally wanted to change the old file name and go with this naming convention but I wasn't sure what to do! Taa muchly!

puppet/modules/chassis/templates/wp-cli.local.yml.erb
BronsonQuick: Doneskis.

PR #810, opened 2020-06-11
Make sure old PHP versions are removed

### LINKED ISSUE #809: Ensure old versions of PHP are removed on a new provision
As a developer, I would like to be able to set my PHP version and ensure that it removes old versions of PHP upon my initial `vagrant up`.

This was reported here: https://github.com/humanmade/altis-local-chassis/issues/91 and I just happened to have hit something similar during my Focal Fossa updates.

BronsonQuick:
Fixes #809.


kadamwhite:
Nice one @BronsonQuick!

PR #816, opened 2020-06-19
Move non-*.local domain warning to run only on "up" and "provision"

kadamwhite:
This is a valid configuration choice and we should not continually yell at users who have opted to go the hostsupdater or manual hosts file modification route.

Moved logic to Vagrantfile to permit it only running on certain actions, rather than every time you SSH. This removes the need for a custom configuration option entirely.

Warning now displays as follows:

image

Original description:

> I do not love the name of this config property, and would accept suggestions for other alternatives.
>
> I also considered suppressing it if `hostsupdater` is `true`, but @BronsonQuick shot that one down :)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -70,8 +70,8 @@ Vagrant.configure("2") do |config|
 					deprecated_extensions << "Please change #{extension} to #{new_extension} in your yaml configuration file.\n"
 				end
 			end
+			trigger.warn = "#{deprecated_extensions}"
 		end
-		trigger.warn = "#{deprecated_extensions}"
 	end
 
 	# Show a warning that non-*.local domains will not automatically resolve.
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -72,6 +72,13 @@ Vagrant.configure("2") do |config|
 			end
 		end
 		trigger.warn = "#{deprecated_extensions}"
+	end
+
+	# Show a warning that non-*.local domains will not automatically resolve.
+	config.trigger.after [ :provision, :up ] do |trigger|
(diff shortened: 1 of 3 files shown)


kadamwhite:
Another alternative path to this solution would be to only show this warning when the VM is initialized, but I do not know how to detect that state from within this file.

kadamwhite:
Moved logic to Vagrantfile to permit it only running on certain actions, rather than every time you SSH. This removes the need for a custom configuration option entirely.

Vagrantfile
```diff
@@ -70,8 +70,15 @@ Vagrant.configure("2") do |config|
 					deprecated_extensions << "Please change #{extension} to #{new_extension} in your yaml configuration file.\n"
 				end
 			end
+			trigger.warn = "#{deprecated_extensions}"
+		end
```
kadamwhite, line 74: This isn't directly related to the stated issue, but moving this warning up a block prevents it from showing an empty

image

if there are no deprecated extensions to show.

Vagrantfile
```diff
@@ -70,8 +70,15 @@ Vagrant.configure("2") do |config|
...
+	end
+
+	# Show a warning that non-*.local domains will not automatically resolve.
+	config.trigger.after [ :provision, :up ] do |trigger|
+		if ! CONF['hosts'][0].include?('.local')
+			trigger.info = "\n\e[33mWARNING: The hosts URL does not contain .local.\nYou will need to edit your hosts file for this URL to resolve.\e[0m"
```
kadamwhite, line 80: This displays as follows:
image

PR #818, opened 2020-07-07
Add Windows IE virtual machine testing guide

kadamwhite:
We put together a draft guide a while back for testing Chassis sites from within Windows VMs. It feels like this is a more appropriate place for that information than our company handbook, so this PR introduces such a guide as part of the Chassis docs site.


kadamwhite:
Also fixes one unrelated syntax issue in code blocks elsewhere in the docs.

PR #821, opened 2020-07-08
Increase RST indent?

### LINKED ISSUE #819: Audit the formatting of the docs
Our docs are generated using RST (reStructuredText) and sometimes it can be picky with formatting. There are a few places in our docs that could use some reformatting so we should audit the docs to fix that up.

kadamwhite:
Fix RST issues so that the new docs guide section renders correctly. Fixes #819.

### DIFF
--- a/docs/guides.rst
+++ b/docs/guides.rst
@@ -279,7 +279,7 @@ When developing on a Mac or Linux host system, you may want to verify your websi
 
    From the command line within your Chassis folder, run ``vagrant ssh -c 'ifconfig'``. This sends a message into your Chassis environment to output the VM's internal network information. What we're looking for is the IP address assigned to the VM, which in our example case looks like this: 
 
-   .. code-block:: txt
+   .. code-block::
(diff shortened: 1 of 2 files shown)


docs/guides.rst
```diff
@@ -277,33 +279,35 @@ When developing on a Mac or Linux host system, you may want to verify your websi
 
 1. **Get your Chassis machine's IP**
 
-   From the command line within your Chassis folder, run ``vagrant ssh -c 'ifconfig'``. This sends a message into your Chassis environment to output the VM's internal network information. What we're looking for is the IP address assigned to the VM, which in our example case looks like this: 
+   From the command line within your Chassis folder, run ``vagrant ssh -c 'ifconfig'``. This sends a message into your Chassis environment to output the VM's internal network information. What we're looking for is the IP address assigned to the VM, which in our example case looks like this:
+
+   ::
```
rmccue, line 284: FYI, this can just be

```
looks like this::
```

(That's where the double-colon syntax originally comes from)

PR #833, opened 2020-09-18
Bump base box to PHP 7.4

### LINKED ISSUE #832: Bump base box to 7.4
It's been a while since I check the [requirements](https://wordpress.org/about/requirements/) on WordPress.org. Turns out they recommend PHP 7.4 so it's time for us to build another base box!

BronsonQuick:
Fixes #832.

I've already built and uploaded the new base box to Vagrant Cloud.

### DIFF
--- a/buildbox.sh
+++ b/buildbox.sh
@@ -52,7 +52,7 @@ echo "\033[0;32mBox minimisation completed.\033[0m"
 vagrant halt
 
 # The version number of the base box.
-VERSION=3.0.5
+VERSION=3.1.0
 
 ## Build the base box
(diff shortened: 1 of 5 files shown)


peterwilsoncc (changes requested):
The PHP version is also mentioned in:

* docs/config.rst
* docs/guides.rst (WPCLI sample, not a huge deal)

That's all I can see outside of the WP folder after a quick search

BronsonQuick:
@peterwilsoncc Good catch Sir! I've updated those references too.

peterwilsoncc (approved):

Issue #852, opened 2021-01-06
Allow installing specific versions of extensions


shadyvb:
1. What operating system do you use?
-

2. What version of Vagrant are you running?
-

3. Are you using VirtualBox or VMWare and which version are you using?
-

4. Do you have a custom YAML file? If so, what does it contain?
   (Please paste in your `chassis.local.yaml` file)
-

5. Do you have any Chassis other extensions installed? If so, what are they?
-

It would be great if we could use specific versions of extensions rather always going for the latest on the main branch. My current use case is being able to use a specific Elastic extension version that has support for Elastic 7.0, which would break the pre-7.0 installs ( unless it is modified to support both, something that isn't straight-forward for me to do ).

I can see the opportunity for Chassis to support an extension name format that would allow for this to happen, eg: `Chassis/chassis_elasticsearch@7.x-compat` or `Chassis/chassis_elasticsearch:7.x-compat`, installation scripts would then understand the format and check out _that branch_ only, eg: `git clone -b 7.x-compat --single-branch {repo_url}`

PR #853, opened 2021-01-07
Allow specifying git branch of extensions

shadyvb:
Resolves https://github.com/Chassis/Chassis/issues/852

I don't really have Ruby experience, but this does seem to work. Happy to update/tweak as needed.

### DIFF
--- a/puppet/chassis.rb
+++ b/puppet/chassis.rb
@@ -253,6 +253,15 @@ module Chassis
 	end
 
 	def self.install_extension(extension)
+		# Check if we have a specific version/branch
+		puts extension
+		if extension =~ /\@/
+			branch = '-b ' + extension.split('@').last
+			extension = extension.gsub(/@.+/, '')
+		else
+			branch = ''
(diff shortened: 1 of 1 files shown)


shadyvb:
Another reason for adoption to this is the `tester` extension being broken when it is updated while project codebases are not, just hit this myself, had to revert the extension for phpunit to work.

BronsonQuick:
@shadyvb Nice work Sir. This looks good to me. Can you please update the documentation to cover this change as well please then I'm happy to test it and merge it. Thank you!

BronsonQuick:
I was having another think about this PR this morning and I am wondering if instead of the `-b branch-name`, it might be better using the commit hash?

BronsonQuick:
@shadyvb Supporting both sounds good to me Sir! I've not been one to maintain branches as it doesn't take long for branches to diverge and not be maintained. I'm happy for us to support both though and other contributors to extensions can maintain the branches.

PR #912, opened 2021-10-11
Suggest updating Chassis box in case of package issues.

shadyvb:
Using old boxes can sometimes cause certificate expired issues while downloading some of the package. Updating to the latest box can solve this issue, so might be worth it to document it in the troubleshooting section.


shadyvb:
@tomjn Unsure what is there that suggests Vagrant boxes are layered or similar to docker images in anyway. Can you clarify what gave you that perception ?

But yeah, we should better clarify that this will involve rebuilding the instance.

PR #938, opened 2021-12-26
Remove our installation of puppet

### LINKED ISSUE #937: Remove the installation of Puppet in `preprovision.sh`
The `bento/ubuntu` box we use now has Puppet 5.5.10 baked into it so we can remove our installation of it.

BronsonQuick:
Fixes #937.

### DIFF
--- a/puppet/preprovision.sh
+++ b/puppet/preprovision.sh
@@ -38,9 +38,6 @@ if [[ ! -f /etc/chassis-updated ]]; then
 	# Update apt
 	sudo apt-get update
 
-	# Install/Upgrade Puppet
-	sudo apt-get -q -y install puppet
-
 	touch /etc/chassis-updated
 fi
 


puppet/preprovision.sh
```diff
@@ -38,9 +38,6 @@ if [[ ! -f /etc/chassis-updated ]]; then
 	# Update apt
 	sudo apt-get update
 
-	# Install/Upgrade Puppet
-	sudo apt-get -q -y install puppet
-
```
tomjn, line 43: could have been wrapped in a conditional so that it only did this when puppet wasn't present, also if no packages are installed is the apt-get update necessary

BronsonQuick: Yeah the conditional is a good idea.

The apt-get update was necessary when we first created this script. I can't remember the reason for it seeing it was so long ago. For now I'll revert this PR.

PR #989, opened 2022-11-09
Use PHP 8 by default

kadamwhite:
PHP 8 should be the default going forward.

Example rationale, WordPress VIP is forcing all remaining PHP 7.4 projects to upgrade within the next week.


BronsonQuick:
Closing in favour of: https://github.com/Chassis/Chassis/issues/1041

PR #995, opened 2022-12-19
Support MariaDB

### LINKED ISSUE #994: Fix support for MariaDB
At some point this extension stopped working: https://github.com/Chassis/mariadb 😞 

I think we should be able to add MariaDB support in Chassis core by adding in some more configuration options [here](https://github.com/Chassis/Chassis/blob/main/puppet/manifests/development.pp#L51-L53). We should be able to pass mysql defaults and get overrides from a YAML file similar to the other mysql [overrides](https://docs.chassis.io/en/latest/config/?highlight=mysql#mysql-configuration).

If we can get that working then we can archive the MariaDB and also update the [WordPress VIP Go](https://github.com/Chassis/vip_go) extension so that MariaDB works out of the box which better matches their environment!

BronsonQuick:
Fixes #994.

We need to test this with no `mysql` options and also with something like the following config in one of the YAML files:
```
mysql:
  mysqld:
   log-error:
     /var/log/mysql/mariadb.log
   pid-file:
     /var/run/mysqld/mysqld.pid
   ssl-disable:
     true
  mysqld_safe:
    log-error:
      /var/log/mysql/mariadb.log
  package_name:
    mariadb-server
```

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -265,6 +265,31 @@ options::
 
 (Again, don't forget to include all lines, and use spaces for indentation.)
 
+---------------------
+MariaDB Configuration
+---------------------
+
+.. py:data:: mysql
+
+You can change the default database flavour to be MariaDB by overridding the default
+MySQL settings::
+
+   mysql:
+       package_name:
+           mariadb-server
+       mysqld:
+           log-error:
+               /var/log/mysql/mariadb.log
+           pid-file:
+               /var/run/mysqld/mysqld.pid
+           ssl-disable:
+               true
+       mysqld_safe:
+           log-error:
+               /var/log/mysql/mariadb.log
+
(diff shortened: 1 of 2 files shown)


BronsonQuick:
Once we've tested this we'll also need to update the documentation for this feature too.

BronsonQuick:
Tested with a default installation:

Site_Health_-_Info__Chassis_Site__WordPress_2022-12-19_17-31-57


Tested with the following YAML:
```
mysql:
  package_name:
    mariadb-server
  mysqld:
   log-error:
     /var/log/mysql/mariadb.log
   pid-file:
     /var/run/mysqld/mysqld.pid
   ssl-disable:
     true
  mysqld_safe:
    log-error:
      /var/log/mysql/mariadb.log
```
Site_Health_-_Info__Chassis_Site__WordPress_2022-12-19_17-39-48

PR #1009, opened 2023-04-07
Add a reinstall script

### LINKED ISSUE #1001: Re-install script to avoid need to destroy.
1. What operating system do you use? macOS 12.6.3
3. What version of Vagrant are you running? 2.3.0
4. Are you using VirtualBox or VMWare and which version are you using? VirtualBox
5. Do you have a custom YAML file? If so, what does it contain? N/A
6. Do you have any Chassis other extensions installed? If so, what are they? N/A

During the development of plugins, it's often necessary to fill the WordPress databases with a bunch of junky content, often using the `wp * generate` commands.

To avoid the need to destroy and re-provision a machine to clean the database, it would be helpful if the install command could be included in a shell script or a custom vagrant command.

The script would drop all the database tables and then reinstall:

```sh
function ask() {
	QUESTION="$1"
	while true; do
		echo
		read -r -p "$QUESTION [y/N] " SHOULD_DO
		SHOULD_DO=$(echo "$SHOULD_DO" | tr '[:upper:]' '[:lower:]')
		case $SHOULD_DO in
			y|yes)
				echo "y"
				return
				;;
			n|no|"")
				return
				;;
			* )
				echo "Please answer yes or no."
				;;
		esac
	done
}

DESTROY=$(ask "Drop all WordPress tables and reinstall?")

if [[ ! -z $DESTROY ]]; then
	wp db clean --yes
	# Chassis generated install command
fi
```

BronsonQuick:
Fixes #1001.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -176,3 +176,21 @@ If you ever want to update Chassis, here's a quick two-step procedure::
 
    # Update WP
    git pull
+
+
+Reinstall
+_________
+
+During the development of plugins, it's often necessary to fill the WordPress
+databases with a bunch of test content, often using the `wp * generate` commands.
+You will often want to reset your WordPress install back to a clean state and not
+want to do a full `vagrant destroy -y && vagrant up` to do so. We have made a script
+that you can run to do this for you.
+
+To run the script, simply run the following command::
+
+	$ vagrant ssh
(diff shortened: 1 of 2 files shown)


peterwilsoncc (commented):
Dandy 🌻

PR #1059, opened 2024-01-01
Change base box for Apple Silicon

### LINKED ISSUE #1058: Alter base box for Apple Silicon chips
We've been using `bento/ubuntu-22.04-arm64` for Apple Silicon chips but we should be using `bento/ubuntu-22.04-arm64`

BronsonQuick:
Fixes #1058.

Monosnap_chassis__vagrant_up_VSchassis__ruby__vagrant_up__14048_2024-01-01_14-02-34
Monosnap_chassis__VSchassis__-fish__14048_2024-01-01_14-06-42
Monosnap_Chassis_Site_2024-01-01_14-07-05

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -127,7 +127,7 @@ Vagrant.configure("2") do |config|
 
 	# The Parallels Provider uses a different naming scheme.
 	config.vm.provider :parallels do |_v, override|
-		# Vagrant currently runs under Rosetta on M1 devices. As a result,
+		# Vagrant currently runs under Rosetta on Apple Silicon devices. As a result,
 		# this seems to be the most reliable way to detect whether or not we're
 		# running under ARM64.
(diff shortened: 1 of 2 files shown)


tomjn:
@BronsonQuick testing a similar change on VVV I noticed provisioning failing a lot due to DNS issues, which I've had in the past. Have you encountered that using Parallels too?

BronsonQuick:
I haven't hit any provisioning issues. Provisioning has always succeeded. I have noticed intermittent Avahi issues. If I do a `vagrant halt` then a `vagrant up` sometimes the `.local` domain won't resolve until after a restart. I haven't been able to replicate that a few times in a row yet though.

tomjn:
hmm i haven't encountered that, rather it fails to resolve the packagist domain or wp.org domains when provisioning things.

Note that it should be enough to restart the avahi daemon rather than the entire VM

PR #1055, opened 2023-12-24
Add WordPress version for automatic update checks

BronsonQuick:
Fixes #1054

PR #1060, opened 2024-01-21
Auto-update workflow alterntive.

peterwilsoncc:
A few differences to #1055.

* If WordPress is out of date, it fails
* Allows the workflow to be manually triggered if we get impatient.
* Separated out each step a little more.
* Uses `wp-includes/version.php`
* Uses PHP
* A little tuning of the PR text.

An example of the PR it creates can be found at https://github.com/peterwilsoncc/Chassis/pull/2

You may wish to squash merge this.


BronsonQuick (approved):
Nice work! TIL `jq` is preinstalled on `ubuntu-latest`.

I also like that you avoided using regex in the version check 🦾  I'll merge this now so we can see it in action for the upcoming release.

PR #1072, opened 2024-04-28
Update apt submodule

BronsonQuick:


Bronson-Akcelo:
I saw this issue today:
```
The specified host network collides with a non-hostonly network!
This will cause your specified IP to be inaccessible. Please change
the IP or name of your host only network so that it no longer matches that of
a bridged or non-hostonly network.

Bridged Network Address: '192.168.56.0'
Host-only Network 'bridge102': '192.168.56.0'
```

I worked around that by adding an ip to my `config.local.yaml` file:
```
ip: 192.168.2.10
```

PR #1078, opened 2024-07-24
Forward SSH port via forwarded_port.

peterwilsoncc:
Modifies how the SSH port forwarding is configured to account for the Parallels Vagrant provider failing to respect `	config.ssh.forward_agent = true`.

~I've tested it with VirtualBox and it appears to work but it seems a little dodgy so I have left this in draft to consider further.~

I think this is in a suitable state, since the original draft push I have limited the changes so they are only applied if the parallels provider is in use. Otherwise the configuration is unchanged.

@BronsonQuick Are you able to test this on both your Silicon and Intel machines and see if it works for you? 

On parallels the SSH address will still use the 10... IP address, so you'll need to use another method to connect via `127.0.0.1:2222` -- I used Sequel Ace.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -125,8 +125,14 @@ Vagrant.configure("2") do |config|
 		config.vm.box = "bento/ubuntu-22.04"
 	end
 
-	# The Parallels Provider uses a different naming scheme.
+	# The Parallels Provider overrides.
 	config.vm.provider :parallels do |_v, override|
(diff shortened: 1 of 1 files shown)


BronsonQuick:
Hey @peterwilsoncc, thanks for the fix. I've tested this today with my Apple Silicon and Intel machines and both are working well.

BronsonQuick (approved):

